PR #8, opened 2018-08-28
Fix load_urdf for urdf files with defined transmissions etc.

ChristophPetzoldt:
Let load_urdf only search for joint and links at top level by setting recursive search in robot.findAll to False.


AlexanderFabisch:
Which problem does it solve? Could you add a unit test that fails with the current version?

AlexanderFabisch:
Ping.

ChristophPetzoldt:
Reopened for pytransform3d at pull request #27 .

PR #10, opened 2018-09-15
Animations

### LINKED ISSUE #9: Make it animatable
Not sure if it's already implemented in some not obvious way, but... I would like to see the library enhanced in a way that plotting the frames/links can be more easily animated. Currently, if I'm not mistaken, it's just `plot`ting, is that right? 
In an animation setup like for example `FuncAnimation` you'd use `set_data` to refresh the plot without having to plot everything again (which is very time consuming). Or does `pytransform` already offer that too? Might be that I've missed something...

AlexanderFabisch:
Fixes #9 

Adds animation feature for transformations and trajectories.

## Example

Generated from `examples/animate_trajectory.py` with

```python
anim.save("trajectory.gif", dpi=80, writer="imagemagick")
```
trajectory

### DIFF
--- a/doc/source/animations.rst
+++ b/doc/source/animations.rst
@@ -1,0 +1,11 @@
+==========
+Animations
+==========
+
+Simple animations of frames are easy to implement with Matplotlib's
+FuncAnimation. The following example will generate a frame that will
+be rotated about the x-axis. The image shown on this page is not
+animated, but you can run the script from the example folder.
+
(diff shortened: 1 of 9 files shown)


AlexanderFabisch:
@HWiese1980 Would that be sufficient for your purpose?

HWiese1980:
Thanks for the effort. Unfortunately not, since I would like to visualize live data.

PR #11, opened 2018-09-22
[WONTMRG] Live Visualization with OpenGL

AlexanderFabisch:
If this is going to be part of pytransform or not has to be decided. It could just remain an advanced example.

### DIFF
--- a/examples/vis_transform.py
+++ b/examples/vis_transform.py
@@ -1,0 +1,187 @@
+import sys
+import numpy as np
+from PyQt4.QtGui import QApplication
+from PyQt4.QtCore import QThread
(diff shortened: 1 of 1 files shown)


AlexanderFabisch:
@HWiese1980 This is a live visualization based on pyqtgraph. All that you have to do is create a new QThread that sends transformation updates with a Qt signal.

HWiese1980:
Yeah, I've been doing something similar with some raw data from the skeleton tracking (no actual use of pytransform yet). PyQtGraph is cool and very easy to use in this regard. Once I've fully switched to pytransform I'll give this a shot. Thanks

PR #17, opened 2018-10-29
Rename pytransform to pytransform3d

### LINKED ISSUE #7: Publish pytransform on PyPI
There is already a package called pytransform: https://pypi.org/project/pytransform/, so we should rename this package. Possible names:

* pytransform3d
* transform3d

(https://packaging.python.org/tutorials/packaging-projects/)

AlexanderFabisch:
As a first step to #7. This will be merged at the end of November. At the same time, the repository will be renamed to pytransform3d and both Travis CI and CircleCI must be updated.


AlexanderFabisch:
@HWiese1980 @ChristophPetzoldt Consider this to be the announcement that the package will be renamed by the end of the month.

HWiese1980:
Alright! Noted... for the moment. I'm sure I will have forgotten it by the end of the... week... But it will inevitably drop on my feet as soon as it becomes relevant for me. 

Does that mean that there will be a pypi package soon like

    pip install pytransform3d

That'd be awesome.

PR #172, opened 2021-12-15
add ability to hide label indicator

alek5k:
The label indicator can look cluttered in certain views and for those not familiar with rgb conventions it may be confusing. I prefer the look of the axes without the indicator, so this PR attempts to make it a little cleaner. I did it through kwargs instead of explicit parameter to be minimally invasive.

### DIFF
--- a/pytransform3d/plot_utils/_artists.py
+++ b/pytransform3d/plot_utils/_artists.py
@@ -20,7 +20,12 @@ class Frame(artist.Artist):
     s : float, optional (default: 1)
         Length of basis vectors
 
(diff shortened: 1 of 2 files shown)


alek5k:
No problem, thanks for the great library!

I've incorporated your suggestions.

AlexanderFabisch:
> No problem, thanks for the great library!

That's nice to hear, thanks!

> I've incorporated your suggestions

I will take a look at it tomorrow in my time zone 🙂

AlexanderFabisch:
Merged to develop. Thanks again!

PR #203, opened 2022-11-29
chore: draft change

teunhoevenaars:
hi @AlexanderFabisch , I appreciate your work on this package! I have been playing around with it and am now building a small development tool that uses this package. 

For my application it would be very helpful if fixed joints can also be identified based on their 'joint_name' and not only via their 'from_name' + 'to_name' combination. Please find attached a draft change. 

Note:
- I did not run the unit tests yet so it's very possible that the current suggestion causes integration issues
- I did not change _get_joint_limits_ since a fixed joint will also have these limits defined.

Curious to hear your thoughts. Perhaps there is a clear reason not to do this. In that case I will solve it locally in my code.

### DIFF
--- a/pytransform3d/urdf.py
+++ b/pytransform3d/urdf.py
@@ -4,6 +4,7 @@ See :doc:`transform_manager` for more information.
 """
 import os
 import numpy as np
+import warnings
 from bs4 import BeautifulSoup
 from .transform_manager import TransformManager
 from .transformations import transform_from, concat
@@ -102,19 +103,21 @@ class UrdfTransformManager(TransformManager):
         from_frame, to_frame, child2parent, axis, limits, joint_type = \
             self._joints[joint_name]
         # this is way faster than np.clip:
-        value = min(max(value, limits[0]), limits[1])
+        value_lim = min(max(value, limits[0]), limits[1])
+        if value_lim != value:
+            warnings.warn(f"Value for {joint_name} is capped by limits")
         if joint_type == "revolute":
             joint_rotation = matrix_from_axis_angle(
-                np.hstack((axis, (value,))))
+                np.hstack((axis, (value_lim,))))
             joint2A = transform_from(
                 joint_rotation, np.zeros(3), strict_check=self.strict_check)
         elif joint_type == "prismatic":
-            joint_offset = value * axis
+            joint_offset = value_lim * axis
             joint2A = transform_from(
                 np.eye(3), joint_offset, strict_check=self.strict_check)
         else:
             assert joint_type == "fixed"
-            print("WARNING: Fixed joint cannot be set")
+            warnings.warn("Trying to set a fixed joint")
             return
         self.add_transform(from_frame, to_frame, concat(
             joint2A, child2parent, strict_check=self.strict_check,
--- a/pytransform3d/urdf.py
+++ b/pytransform3d/urdf.py
@@ -103,16 +103,14 @@ class UrdfTransformManager(TransformManager):
         from_frame, to_frame, child2parent, axis, limits, joint_type = \
             self._joints[joint_name]
         # this is way faster than np.clip:
-        value_lim = min(max(value, limits[0]), limits[1])
-        if value_lim != value:
-            warnings.warn(f"Value for {joint_name} is capped by limits")
+        value = min(max(value, limits[0]), limits[1])
         if joint_type == "revolute":
             joint_rotation = matrix_from_axis_angle(
-                np.hstack((axis, (value_lim,))))
+                np.hstack((axis, (value,))))
             joint2A = transform_from(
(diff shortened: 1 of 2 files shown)


pytransform3d/urdf.py
```diff
@@ -108,11 +108,14 @@ def set_joint(self, joint_name, value):
...
                 joint_rotation, np.zeros(3), strict_check=self.strict_check)
-        else:
-            assert joint_type == "prismatic"
+        elif joint_type == "prismatic":
             joint_offset = value * axis
             joint2A = transform_from(
                 np.eye(3), joint_offset, strict_check=self.strict_check)
+        else:
+            assert joint_type == "fixed"
+            print("WARNING: Fixed joint cannot be set")
```
AlexanderFabisch, line 117: I would raise an error or print a warning here.

teunhoevenaars: Thanks for the quick feedback :). I originally thought to raise an error, but wanted to first see if this conflicts with the unit tests. Good to hear you have the same idea

AlexanderFabisch:
Looks reasonable to me. The only concern that I have is that I don't know what `get_joint_limits` should return for this joint. Should it raise an error?

pytransform3d/urdf.py
AlexanderFabisch: I didn't run the tests, but I see no serious problems. We would only need additional unit tests for the new feature.

AlexanderFabisch:
btw. @teunhoevenaars , maybe this fits your use case better: you can also directly call the function [parse_urdf](https://github.com/teunhoevenaars/pytransform3d/blob/feature/fixed_joints_in_urdf/pytransform3d/urdf.py#L289) to get all information that is currently parsed from the URDF

teunhoevenaars:
thanks for the suggestion. However, I really like the fact that the _UrdfTransformManager_ contains all the information in one object. With this PR, I could just loop over __joints.items()_ (which is a protected attribute, but very useful for me) and perform work on them based on identification by name (instead of _to_frame_ and _from_frame_)

pytransform3d/urdf.py
teunhoevenaars: - existing tests do not give issues.

> Hi,
> 
> thanks for the contribution. Looks reasonable to me. The only concern that I have is that I don't know what `get_joint_limits` should return for this joint. Should it raise an error?

I have currently implemented this such that 'fixed' joints are hard-coded with limits (0, 0). The limits therefore enforce the "fixedness" of the joint. Also, this ensures that this joint cannot be changed by _set_joint_ because it will cap the input to these limits.

AlexanderFabisch:
Although accessing protected attributes is not recommended, in this case I don't expect the interface to change in the near future, so I think that should be safe, too.

pytransform3d/urdf.py
AlexanderFabisch: I think I'd rather see a warning somewhere though. Because it doesn't make sense to change the joint angle of a fixed joint.

teunhoevenaars: found it, using _warnings_

pytransform3d/urdf.py
```diff
@@ -102,17 +103,22 @@ def set_joint(self, joint_name, value):
         from_frame, to_frame, child2parent, axis, limits, joint_type = \
             self._joints[joint_name]
         # this is way faster than np.clip:
-        value = min(max(value, limits[0]), limits[1])
+        value_lim = min(max(value, limits[0]), limits[1])
+        if value_lim != value:
+            warnings.warn(f"Value for {joint_name} is capped by limits")
```
AlexanderFabisch, line 108: This changes existing behavior, which I would like to avoid. Values for other joint types were just clipped silently before. I think that's generally accepted behavior. If it's not documented in the function's docstring, we could think about documenting this behavior.

AlexanderFabisch:
Apart from this comment I think it can be merged. I would just test the test coverage before that.

pytransform3d/urdf.py
teunhoevenaars: OK, understood, I've undone these changes.

teunhoevenaars:
Great to hear Alexander! I understand that testing the test coverage is something you will do, let me know if I've misunderstood or there is something I can do

AlexanderFabisch:
Yes, exactly. Merged to develop branch. Thanks for your contribution!

teunhoevenaars:
thanks @AlexanderFabisch! My aim is to remain involved, so hopefully we'll be in touch again soon enough. For my info; approximately when is the next Release scheduled / foreseen?

AlexanderFabisch:
That sounds good. I am always happy to get some support!

There is no fixed release schedule. I usually update the version number after adding a few new features.

PR #208, opened 2023-01-01
Conversion from quaternions to arbitrary Euler angles

### LINKED ISSUE #207: Implement direct conversion between quaternions and euler angles in any sequence
I recently published a [paper (open access)](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0276302) on a direct conversion between quaternions and euler angles in any sequence, that is a lot simpler to implement, faster to compute and does not need an intermediate step converting it to a matrix.

Since its publication, it has already been merged into [Scipy](https://github.com/scipy/scipy/pull/17392#issuecomment-1341593229), replacing the existing method, and it is also now part of [Sympy](https://github.com/sympy/sympy/pull/24358).

Since the original Scipy `Rotation.as_euler` method was used as an inspiration for `_general_intrinsic_euler_from_active_matrix`, the new direct method could be a nice addition to this library.

AlexanderFabisch:
Fixes #207 

TODO

- [x] numerical safety as in [[1]](https://osf.io/sedhn) and [[2]](https://github.com/sympy/sympy/pull/24358/files)
- [x] make general interface instead of 24 different functions
- [x] test coverage
- [x] test edge cases


AlexanderFabisch:
@evbernardes Since this is your algorithm, would you have time for a review?

pytransform3d/rotations/_conversions.py
```diff
@@ -1383,6 +1383,122 @@ def extrinsic_euler_zxy_from_active_matrix(R, strict_check=True):
...
+    d = q[k] * sign
+
+    if not proper_euler:
+        a, b, c, d = a - c, b + d, c + a, d - b
+
+    # Equation 35
+    angle_j = math.acos(
+        2.0 * (a * a + b * b) / (a * a + b * b + c * c + d * d) - 1.0)
+    # or Equation 34; TODO which one should be used and why?
+    #angle_j = 2.0 * math.atan2(math.sqrt(c * c + d * d),
```
evbernardes, line 1451: This is a good question. 
I had opted for Equation 35 but ended up changing to Equation 34 during the Scipy implementation.
Apparently, `acos` and `asin` implementations might have slight numerical problems near `zero`, making it preferable to use `atan2` whenever possible.
I would implement this as:
```
2 * math.atan2(math.hypot(c, d), math.hypot(a, b))
```

pytransform3d/rotations/_conversions.py
```diff
@@ -1383,6 +1383,122 @@ def extrinsic_euler_zxy_from_active_matrix(R, strict_check=True):
...
+def euler_from_quaternion(q, i, j, k, extrinsic):
+    """General method to extract any Euler angles from quaternions.
+
+    Parameters
+    ----------
+    q : array-like, shape (4,)
+        Unit quaternion to represent rotation: (w, x, y, z)
+
+    i : int
+        Axis of the first rotation angle. A number between 0 and 2.
```
evbernardes, line 1395: Aren't axes more usually defined as numbers between `1` and `3`?

pytransform3d/rotations/_conversions.py
```diff
@@ -1383,6 +1383,122 @@ def extrinsic_euler_zxy_from_active_matrix(R, strict_check=True):
...
+    a = q[0]
+    b = q[i]
+    c = q[j]
+    d = q[k] * sign
+
+    if not proper_euler:
+        a, b, c, d = a - c, b + d, c + a, d - b
+
+    # Equation 35
+    angle_j = math.acos(
```
evbernardes, line 1448: Out of curiosity, is there any reason to use the `math` module since you're already using `numpy`?

evbernardes (commented):
If the tests are passing, then everything looks good!

As a new addition, I'd personally make a corresponding `quaternion_from_euler` function too though, not only for consistency but also for easier testing (since this operation is trivial.)

pytransform3d/rotations/_conversions.py
```diff
@@ -1383,6 +1383,122 @@ def extrinsic_euler_zxy_from_active_matrix(R, strict_check=True):
...
+
+    # Proper Euler angles rotate about the same axis in the first and last
+    # rotation. If this is not the case, they are called Cardan or Tait-Bryan
+    # angles and have to be handled differently.
+    proper_euler = i == k
+    if proper_euler:
+        k = 6 - i - j
+
+    sign = (i - j) * (j - k) * (k - i) // 2
+    a = q[0]
```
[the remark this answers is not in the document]
AlexanderFabisch: Scalar first is the default in this library. It is documented, but not really possible to check, since it uses plain numpy arrays.

pytransform3d/rotations/_conversions.py
AlexanderFabisch: This is consistent with [matrix_from_angle](https://dfki-ric.github.io/pytransform3d/_apidoc/pytransform3d.rotations.active_matrix_from_angle.html#pytransform3d.rotations.active_matrix_from_angle).

pytransform3d/rotations/_conversions.py
AlexanderFabisch: It's faster for scalars. When operating on arrays, numpy is faster though.

```python
In [1]: import numpy as np

In [2]: import math

In [3]: %timeit np.arccos(np.random.rand())
1.34 µs ± 12.6 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [4]: %timeit math.acos(np.random.rand())
384 ns ± 5.94 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit np.arctan2(np.random.rand(), np.random.rand())
1.88 µs ± 21.4 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [6]: %timeit math.atan2(np.random.rand(), np.random.rand())
689 ns ± 9.63 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

AlexanderFabisch:
> As a new addition, I'd personally make a corresponding quaternion_from_euler function too though, not only for consistency but also for easier testing (since this operation is trivial.)

That's maybe another pull request. I'll open another issue for it.

Thanks for the review and the suggestion!

PR #209, opened 2023-01-02
Fix bug in plot_cylinder (#206)

### LINKED ISSUE #206: Bug in plot_cylinder function.
I was recently using plot_cylinder to draw joints for a robot. I found that an `RuntimeWarning` occurs when the cylindrical z-axis is exactly in the negative direction of the x-axis. Also the cylinder does not appear in axis.

image

Here is my code:

```python
import numpy as np
import matplotlib.pyplot as plt
from pytransform3d.transformations import plot_transform
from pytransform3d.plot_utils import plot_cylinder

a2b = np.array([[0, 0, -1., 0],
                [1, 0, -0, 0],
                [0, -1, 0, 0],
                [0, 0, 0, 1]])

ax = plot_cylinder(length=1.0, radius=0.3, thickness=0.1, A2B=a2b, wireframe=False, alpha=0.2)
plot_transform(ax=ax, A2B=a2b, s=0.3, lw=3)
plt.show()
```

Here is the output:

```
...pytransform3d\pytransform3d\plot_utils\_plot_functions.py:216: RuntimeWarning: invalid value encountered in divide
  n1 /= np.linalg.norm(n1)
```

I tried to change the code in line 212 of file `pytransform3d\plot_utils\_plot_functions.py` and now the program works as I expected.

image

Here are the changes I made:

```
# original code of pytransform3d\plot_utils\_plot_functions.py at line 212
    not_axis = np.array([1, 0, 0])
    if (axis == not_axis).all():
        not_axis = np.array([0, 1, 0])

# modified code
    not_axis = np.array([1, 0, 0])
    if np.allclose(axis, not_axis) or np.allclose(-axis, not_axis):
        not_axis = np.array([0, 1, 0])
```

chinaheyu:
Fix #206

### DIFF
--- a/pytransform3d/plot_utils/_plot_functions.py
+++ b/pytransform3d/plot_utils/_plot_functions.py
@@ -209,7 +209,7 @@ def plot_cylinder(ax=None, length=1.0, radius=1.0, thickness=0.0,
     axis /= length
 
     not_axis = np.array([1, 0, 0])
-    if (axis == not_axis).all():
+    if np.allclose(axis, not_axis) or np.allclose(-axis, not_axis):
         not_axis = np.array([0, 1, 0])
 
     n1 = np.cross(axis, not_axis)


AlexanderFabisch:
Thanks for your contribution!

Issue #54, opened 2020-06-19
Increase tolerance threshold for matrix check


theRealSuperMario:
Hi, I would like to use your library for my project but I get a value error in the line below.

I am pretty sure that my matrix is correct as it represents the extrinsics of a camera.
When I increase the a-tol value `np.allclose`, it works just fine

![image](https://user-images.githubusercontent.com/6517465/85123514-b0086e80-b228-11ea-965d-e94d51de9868.png)

So my suggestion would be to:
1. change the raise to a warning so that one can keep working without hacking around in the library
2. consider pulling up an a-tol flag to every call of the method. However, this would be a pretty large rafactoring.
3. Alternatively, one could try to find a context manager that switches the precision of `np.allclose`, however, I was not able to find this.

https://github.com/rock-learning/pytransform3d/blob/f4c5c01132df31c0c81962ee3857e670ff026699/pytransform3d/rotations.py#L251

I am happy to hear back on suggestions. Apart from that, I think the library is pretty nice to work with and goes well with http://matthew-brett.github.io/transforms3d/

theRealSuperMario:
I added a possible fix here:

https://github.com/theRealSuperMario/pytransform3d/tree/numerical_precision

I could make a PR if desired

AlexanderFabisch:
Hi,

thanks for the feedback. I would need more information.

* I assume the rotation matrix was not computed in pytransform3d. How did you obtain it?
* Can you tell me the exact values to reproduce the error?
* The second check (for the determinant) was introduced only recently. Did you have problems with that?

I agree that raising an error every time something is slightly incorrect is not the most convenient behavior. Numpy has a global function to define error handling: [numpy.seterr](https://numpy.org/doc/stable/reference/generated/numpy.seterr.html). Maybe pytransform3d should have something similar. In addition, I could raise the error threshold a bit, but I'd like to see the exact numbers first.

> I think the library is pretty nice to work with

Good to hear that!

theRealSuperMario:
1. It is from a public dataset (human36M)
2. 
```
0:array([-9.15361835e-01,  4.01808328e-01,  2.57475872e-02, -3.46050839e+02])
1:array([ 5.15480570e-02,  1.80374088e-01, -9.82246499e-01,  5.46980731e+02])
2:array([-3.99318925e-01, -8.97783496e-01, -1.85819250e-01,  5.47448099e+03])
```
Only the [:3,:3] slice is rotation

3. The determinant was not a problem

[...]

PR #231, opened 2023-03-06
Adds missing strict_check argument in camera.py

Alpe6825:
Adds missing strict_check argument. (see #54)

Fixes problem that occurs when plotting cameras. cam2world will be checked in line 266 and inside of transform in line 278. For transform the strickt_check argument was missing.


AlexanderFabisch:
Thanks for the contribution!

AlexanderFabisch:
I merged it to the branch develop-3 and it will be part of release 3.0.0.

PR #247, opened 2023-04-25
Fix and test edge case for matrix conversion

tungalbert99:
There's a slight edge case I found due to numerical instability where the matrix conversion will error for arccosine if the value is 1.0 + epsilon and provide a NaN value. It will trigger specifically the code path in lines 995-996 but I have not found an orientation that triggers 997-998. If you want, I can remove that code path or add a corresponding test.

### DIFF
--- a/pytransform3d/test/test_rotations.py
+++ b/pytransform3d/test/test_rotations.py
@@ -2225,3 +2225,9 @@ def test_jacobian_so3():
 
         J_inv_J = np.dot(J_inv, J)
         assert_array_almost_equal(J_inv_J, np.eye(3))
+
+def test_euler_from_quaternion_edge_case():
+    quaternion = [0.57114154, -0.41689009, -0.57114154, -0.41689009]
+    matrix = pr.matrix_from_quaternion(quaternion)
+    euler_xyz = pr.extrinsic_euler_xyz_from_active_matrix(matrix)
+    assert not np.isnan(euler_xyz).all()
--- a/pytransform3d/test/test_rotations.py
+++ b/pytransform3d/test/test_rotations.py
@@ -2226,8 +2226,10 @@ def test_jacobian_so3():
         J_inv_J = np.dot(J_inv, J)
         assert_array_almost_equal(J_inv_J, np.eye(3))
 
+
 def test_euler_from_quaternion_edge_case():
     quaternion = [0.57114154, -0.41689009, -0.57114154, -0.41689009]
     matrix = pr.matrix_from_quaternion(quaternion)
     euler_xyz = pr.extrinsic_euler_xyz_from_active_matrix(matrix)
     assert not np.isnan(euler_xyz).all()
+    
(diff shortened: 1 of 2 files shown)


pytransform3d/test/test_rotations.py
```diff
@@ -2225,3 +2225,9 @@ def test_jacobian_so3():
 
         J_inv_J = np.dot(J_inv, J)
         assert_array_almost_equal(J_inv_J, np.eye(3))
+
+def test_euler_from_quaternion_edge_case():
+    quaternion = [0.57114154, -0.41689009, -0.57114154, -0.41689009]
+    matrix = pr.matrix_from_quaternion(quaternion)
+    euler_xyz = pr.extrinsic_euler_xyz_from_active_matrix(matrix)
+    assert not np.isnan(euler_xyz).all()
```
AlexanderFabisch, line 2233: PEP8: please add another newline at the end of the file

AlexanderFabisch:
Hi @tungalbert99 ,

thanks for pointing this out! I added two comments on the format. Then we can merge it to master with the next release of the library.

pytransform3d/test/test_rotations.py
tungalbert99: done! sorry my formatter is very different so I had to save without it

pytransform3d/test/test_rotations.py
```diff
@@ -2225,3 +2225,9 @@ def test_jacobian_so3():
 
         J_inv_J = np.dot(J_inv, J)
         assert_array_almost_equal(J_inv_J, np.eye(3))
+
```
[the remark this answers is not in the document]
tungalbert99: done!

tungalbert99 (commented):
fixed new line

AlexanderFabisch:
Thanks. I would like to keep 100% coverage. Do you have an idea of how we could find a test case for the missing branch?

AlexanderFabisch:
@tungalbert99 I looked at the fix again and I changed it to [this solution](https://github.com/dfki-ric/pytransform3d/blob/e5deab7e7c391008fbb79c2a2c3f592888fca3cf/pytransform3d/rotations/_conversions.py#L995). Was there a reason why you set values close to 0 to exactly 0? It does not seem to make sense in my opinion. Forcing an argument of arccos to be within [-1, 1] does make sense to me intuitively, of course.

tungalbert99:
Sorry, I think I mistaken the domain for arccosine when I wrote this. The reason I didn't clip the argument was that if it exceeded the bounds of [-1,1] by more than epsilon (absolute diff), then it's likely that an invalid rotation was passed in? I'm not sure how valid that conjecture is though.

AlexanderFabisch:
We check the correctness of the rotation matrix in the first line of the function already. So I will just clip the values. Thanks for clarification!

tungalbert99:
Awesome! Thanks for maintaining an incredibly useful library

AlexanderFabisch:
Thanks a lot! :)

PR #256, opened 2023-07-14
[WIP] Refactor TransformManager and introduce TemporalTransformManager

AlexanderFabisch:


pytransform3d/transform_manager.pyi
```diff
@@ -1,15 +1,17 @@
...
 from typing import Dict, Tuple, List, Union, Set, Hashable, Any
 
 
-class TransformManager(object):
+class TransformGraphBase(abc.ABC):
     strict_check: bool
     check: bool
-    transforms: Dict[Tuple[Hashable, Hashable], np.ndarray]
     nodes: List[Hashable]
+    transforms: Dict[Tuple[Hashable, Hashable], np.ndarray]
```
kopytjuk, line 14: This would force the future `TemporalTransformManager` to use `np.ndarray`s to store timeseries of transformation. I rather suggest the following:

```python
from typing import Protocol

class Transform(Protocol):

    def get_matrix(self, time: Optional[float]):
        ...
```

```suggestion
    transforms: Dict[Tuple[Hashable, Hashable], Transform]
```

pytransform3d/transform_manager.pyi
```diff
@@ -18,20 +20,50 @@ class TransformManager(object):
...
+
+    def _recompute_shortest_path(self): ...
 
     def remove_transform(
             self, from_frame: Hashable,
             to_frame: Hashable) -> "TransformManager": ...
 
-    def has_frame(self, frame: Hashable) -> bool: ...
+    def get_transform(
+            self, from_frame: Hashable, to_frame: Hashable) -> Any: ...
```
kopytjuk, line 37:
```suggestion
    def get_transform(
            self, from_frame: Hashable, to_frame: Hashable) -> Transform: ...
```

AlexanderFabisch: We could remove this from the base class if required

pytransform3d/transform_manager.pyi
AlexanderFabisch: This is a property now

pytransform3d/transform_manager/_transform_graph_base.pyi
```diff
@@ -0,0 +1,66 @@
...
+
+    @property
+    @abc.abstractmethod
+    def transforms(self) -> Dict[Tuple[Hashable, Hashable], np.ndarray]: ...
+
+    @abc.abstractmethod
+    def _check_transform(self, A2B: Any) -> Any: ...
+
+    @abc.abstractmethod
+    def _invert_transform(self, A2B: Any) -> Any: ...
```
kopytjuk, line 31: If the `self.transforms` property always returns numpy-array, do we need `_invert_transform` method? We can call the utility function directly in `get_transform()` method. This would be OK with Temporal Manager too

pytransform3d/transform_manager/_transform_graph_base.pyi
```diff
@@ -0,0 +1,66 @@
...
+
+    @abc.abstractmethod
+    def _path_transform(self, path: List[Hashable]) -> Any: ...
```
kopytjuk, line 34: If the `self.transforms` property always returns numpy-array, we can implement this method in the base class. Our "unit of operation" is a 4X4 matrix anyway...

pytransform3d/transform_manager/_transform_graph_base.pyi
```diff
@@ -0,0 +1,66 @@
...
+
+    @abc.abstractmethod
+    def _transform_available(self, key: Tuple[Hashable, Hashable]) -> bool: ...
```
kopytjuk, line 37: I think it is not needed, we can check for a key in `transforms` anyway ...

pytransform3d/transform_manager/_transform_graph_base.pyi
```diff
@@ -0,0 +1,66 @@
...
+
+    @abc.abstractmethod
+    def _set_transform(self, key: Tuple[Hashable, Hashable], A2B: Any): ...
+
+    @abc.abstractmethod
+    def _get_transform(self, key: Tuple[Hashable, Hashable]) -> Any: ...
```
kopytjuk, line 43: If the `self.transforms` property always returns numpy-array, we do not need this method I think ...

pytransform3d/transform_manager/_transform_graph_base.py
```diff
@@ -0,0 +1,275 @@
...
+            them
+        """
+        if self.check:
+            if from_frame not in self.nodes:
+                raise KeyError("Unknown frame '%s'" % from_frame)
+            if to_frame not in self.nodes:
+                raise KeyError("Unknown frame '%s'" % to_frame)
+
+        if self._transform_available((from_frame, to_frame)):
+            return self._get_transform((from_frame, to_frame))
```
kopytjuk, line 208: Consider just using the `self.transforms` property ...

kopytjuk:
Hey @AlexanderFabisch ,

As you can see in https://github.com/dfki-ric/pytransform3d/pull/257, the state of this PR is fine for the new stateful TemporalManager (great idea btw!).

Two small (non-blocking) remarks:

- You still can "tweak" around a bit, when you move `_path_transform` to the base class here.
- Do we still need so many abstract interfaces such as `_get_transform`, `_transform_available `, `_invert_transform` - imho we can just use the `transforms` property

See you around :)

AlexanderFabisch:
@kopytjuk 

So here are some open points:

- [ ] Add TemporalTransformManager to API documentation
- [ ] Update tutorial about temporal transformations
- [ ] Discussion: should we include another subclass of TimeVaryingTransformation for dynamic transformations that allows interpolation?

For the last point: I think we could do it. We could also do this based on pandas, but only with pandas as an optional dependency. If it is not available, this feature is just not available.

For the documentation I'm not sure how it should be changed. @kopytjuk you are the expert in this domain, would you take a look?

kopytjuk:
Should we split those two open points in separate PRs?

Currently the changes from this PR affect the "backend" of the lib - after we merge this one into develop we will extend the documentation.

Imho it is easier to digest smaller changes than a huge PR...

AlexanderFabisch:
Yes, I just did that.

PR #257, opened 2023-07-16
Implement TemporalTrafoManager on top of base-class

kopytjuk:


pytransform3d/temporal_transform_manager.py
```diff
@@ -0,0 +1,40 @@
...
+        return A2B
+    
+    def _invert_transform(self, A2B):
+        """Invert rigid transformation stored in the tree."""
+        return A2B.invert()
+    
+    def _path_transform(self, path):
+        A2B = NumpyTransform(np.eye(4))
+        for from_f, to_f in zip(path[:-1], path[1:]):
+            A2B = A2B.concat(self.get_transform(from_f, to_f))
```
kopytjuk, line 39: `concat` operation has to deal with two transformation with different time-bases ...

AlexanderFabisch: I would extend _path_transform to include the additional arguments from `get_transform` such as `time=x` so that you only have to interpolate at time `x`.

pytransform3d/transform_manager/_transform_manager.pyi
```diff
@@ -17,10 +17,6 @@ class TransformManager(TransformGraphBase):
     @property
     def transforms(self) -> Dict[Tuple[Hashable, Hashable], np.ndarray]: ...
 
-    def _check_transform(self, A2B: npt.ArrayLike) -> np.ndarray: ...
```
kopytjuk, line 20: They are now part of the base class

pytransform3d/transform_manager/_temporal_transform_manager.py
kopytjuk, line 1: I will add the `_temporal_transform_manager.pyi` after we agreed on the interfaces and the contents of the base class

pytransform3d/transform_manager/_transform_graph_base.pyi
```diff
@@ -16,22 +16,18 @@ class TransformGraphBase(abc.ABC):
     dist: np.ndarray
     predecessors: np.ndarray
     _cached_shortest_paths: Dict[Tuple[int, int], List[Hashable]]
+    _transforms = Dict[Tuple[Hashable, Hashable], Any]
```
kopytjuk, line 19: We can put here what we want, and the get_transforms should call the interface/protocol ... i.e. deal with numpy arrays or `TimeTransform` instances

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -0,0 +1,86 @@
...
+
+from ._transform_graph_base import TransformGraphBase
+from ..transformations import (check_transform, invert_transform, concat)
+
+
+class TimeTransform(abc.ABC):
+    """Base class passed to TemporalTransformManager.add_transform() to deal with time.
+    """
+    
+    @abc.abstractmethod
```
kopytjuk, line 14: Alternatively we could describe the object passed as a `Protocol`, by that way, clients do not have to subclass `TimeTransform` and can just implement `as_matrix` method ... which would decouple code ...

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -0,0 +1,86 @@
...
+            them
+        """
+        if self.check:
+            if from_frame not in self.nodes:
+                raise KeyError("Unknown frame '%s'" % from_frame)
+            if to_frame not in self.nodes:
+                raise KeyError("Unknown frame '%s'" % to_frame)
+
+        if (from_frame, to_frame) in self.transforms:
+            return self._transforms[(from_frame, to_frame).as_matrix(time)]
```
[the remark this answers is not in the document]
AlexanderFabisch: You have to close with `]` before you can call `as_matrix`.

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -0,0 +1,86 @@
...
+            raise KeyError("Cannot compute path from frame '%s' to "
+                           "frame '%s'." % (from_frame, to_frame))
+
+        path = self._shortest_path(i, j)
+        return self._path_transform(path, time)
+    
+    @property
+    def transforms(self):
+        """Rigid transformations between nodes."""
+        return self._transforms
```
AlexanderFabisch, line 72: When we have the state variable, we could also return all transformations at that time here.

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -0,0 +1,86 @@
...
+
+    def _check_transform(self, A2B):
```
AlexanderFabisch, line 74: You should override this function to check all transformations.

I would suggest to extend the interface of `TimeTransform` with a function `check_transform` that does exactly this for all transformations in the sequence.

pytransform3d/transform_manager/_transform_graph_base.py
```diff
@@ -37,6 +39,7 @@ def __init__(self, strict_check=True, check=True):
         # distance matrix (distance is the number of transformations)
         self.dist = np.empty(0)
         self.predecessors = np.empty(0, dtype=np.int32)
+        self._transforms = {}
```
AlexanderFabisch, line 42: As the type of this dictionary is different between the normal and temporal TransformManager, I wouldn't include it in the base class.

pytransform3d/transform_manager/_transform_graph_base.py
```diff
@@ -147,8 +151,8 @@ def remove_transform(self, from_frame, to_frame):
             This object for chaining
         """
         transform_key = (from_frame, to_frame)
-        if transform_key in self.transforms:
-            del self.transforms[transform_key]
+        if transform_key in self._transforms:
```
AlexanderFabisch, line 154: Good catch, I will also change that in the other PR

pytransform3d/transform_manager/_temporal_transform_manager.py
kopytjuk: Done

pytransform3d/transform_manager/_temporal_transform_manager.py
kopytjuk: This is an amazing idea!

kopytjuk:
Hey @AlexanderFabisch 

your idea with a state variable holding the current time was amazing!

So the interpolation is conducted within the `transforms` property method calling the `as_matrix(time)` interface there. 

We could even remove the `_get_transform` methods, and access `transforms` property directly in the base class!

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -0,0 +1,81 @@
...
+        return self._current_time
+
+    def set_time(self, time):
+        self._current_time = time
+    
+    @property
+    def transforms(self):
+        """Rigid transformations between nodes."""
+        return {tf_direction: transform.as_matrix(self.current_time) for
+                tf_direction, transform in self._transforms.items()}
```
kopytjuk, line 62: This is the key change, basically... 

We could even think about [cached properties](https://docs.python.org/3.8/library/functools.html#functools.cached_property) (but they unfortunately are from >=3.8). This would avoid re-computing interpolated transforms ...

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -0,0 +1,91 @@
...
+    
+    def get_transform_in_time(self, from_frame, to_frame, time):
```
kopytjuk, line 64: Nicer API for users

pytransform3d/transform_manager/_temporal_transform_manager.py
AlexanderFabisch: It would decouple code, but errors would still occur only at runtime. At least with this approach you know that you forgot to implement a required function.

pytransform3d/transform_manager/_temporal_transform_manager.py
AlexanderFabisch: But I think it wouldn't be so easy to recompute the cached property when we change the current time, right?

AlexanderFabisch (changes requested):
Everything looks very good now.

I'd like to merge this PR to refactor/transform_manager soon.

What is missing to close this first step is the following:

- [x] there are some comments above
- [x] proper testing to reach 100% coverage
- [ ] separation of type hints in .pyi file

After that we have to address some other issues:

- [ ] update documentation (tutorial part)
- [ ] discuss inclusion of dynamic transformation (documentation, additional optional dependencies?)

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -0,0 +1,141 @@
...
+        super(TemporalTransformManager, self).__init__(strict_check, check)
+        self._transforms: Dict[Tuple[Hashable, Hashable], TimeVaryingTransform] = {}
+        self._current_time: float = 0.0
+
+    @property
+    def current_time(self) -> float:
+        """Current time at which we evaluate transformations."""
+        return self._current_time
+
+    def set_time(self, time: float):
```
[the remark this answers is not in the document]
kopytjuk: Adapted.

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -0,0 +1,141 @@
...
+        self._transforms[key] = A2B
+
+    def _get_transform(self, key):
+        return self._transforms[key].as_matrix(self._current_time)
+
+    def _del_transform(self, key):
+        del self._transforms[key]
+
+    def _check_transform(self, A2B):
+        """Check validity of rigid transformation."""
```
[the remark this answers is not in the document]
kopytjuk: Will remove

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -0,0 +1,141 @@
...
+        previous_time = self._current_time
+        self.set_time(time)
+
+        A2B = self.get_transform(from_frame, to_frame)
+
+        # revert internal state
+        self.set_time(previous_time)
+        return A2B
+
+    def _transform_available(self, key):
```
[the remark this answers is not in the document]
kopytjuk: Okay you are right, let's keep that as it is

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -0,0 +1,141 @@
...
+
+from ._transform_graph_base import TransformGraphBase
+from ..transformations import check_transform
+
+
+class TimeVaryingTransform(abc.ABC):
+    """Time-varying rigid transformation."""
+
+    @abc.abstractmethod
+    def as_matrix(self, time) -> np.ndarray:
```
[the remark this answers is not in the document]
kopytjuk: Added a pyi file

pytransform3d/transform_manager/_transform_graph_base.py
```diff
@@ -49,13 +51,18 @@ def transforms(self):
     def _check_transform(self, A2B):
         """Check validity of rigid transformation."""
 
-    @abc.abstractmethod
-    def _invert_transform(self, A2B):
+    def _invert_transform(self, A2B_matrix):
```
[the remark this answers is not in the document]
kopytjuk: Removed

pytransform3d/test/test_temporal_transform_manager.py
kopytjuk, line 1: Now we have 100% test coverage

AlexanderFabisch (changes requested):
Apart from these last few comments it looks good!

pytransform3d/transform_manager/_temporal_transform_manager.pyi
```diff
@@ -0,0 +1,52 @@
...
+import numpy.typing as npt
+
+from ._transform_graph_base import TransformGraphBase
+
+
+class TimeVaryingTransform(abc.ABC):
+    """Time-varying rigid transformation."""
+
+    @abc.abstractmethod
+    def as_matrix(self, time) -> np.ndarray: ...
```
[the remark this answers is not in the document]
kopytjuk: Updated.

pytransform3d/transform_manager/_temporal_transform_manager.pyi
```diff
@@ -0,0 +1,52 @@
...
+from typing import Dict, Tuple, Hashable, Any
+
+import numpy as np
+import numpy.typing as npt
+
+from ._transform_graph_base import TransformGraphBase
+
+
+class TimeVaryingTransform(abc.ABC):
+    """Time-varying rigid transformation."""
```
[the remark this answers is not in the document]
kopytjuk: Removed it in the stub

AlexanderFabisch (approved):

PR #266, opened 2023-08-06
Fix arccos numerical issue in axis_angle_from_two_directions

danielpmorton:
There was an issue with arccos similar to https://github.com/dfki-ric/pytransform3d/pull/247, so I've clamped the bounds similar to how you've done this in other areas of the code with arccos

All other arccos instances in the `_conversions` file are clamped, except for this and one other place - in `axis_angle_from_quaternion`. Should I clamp that one as well?

### DIFF
--- a/pytransform3d/batch_rotations.py
+++ b/pytransform3d/batch_rotations.py
@@ -302,7 +302,7 @@ def axis_angles_from_matrices(Rs, traces=None, out=None):
             # out[False, n] = value will not assign value to out[n]
             traces = traces[0]
 
-    angles = np.arccos((traces - 1.0) / 2.0)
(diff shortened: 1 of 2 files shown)


AlexanderFabisch:
Hi @danielpmorton ,

thanks for this contribution.

> All other arccos instances in the _conversions file are clamped, except for this and one other place - in axis_angle_from_quaternion. Should I clamp that one as well?

Yes, I think it makes sense to clamp it in `axis_angle_from_quaternion` as well.

danielpmorton:
Ok, just updated that function, and another instance I found in `batch_rotations`

AlexanderFabisch:
Merged to develop, thanks a lot!

PR #300, opened 2024-10-07
Add remove_frame Method to Manage Graph Size

clungzta:
A `remove_frame` method to `TransformGraphBase`, allowing for the removal of nodes (frames) that are no longer in use. Helping to keep the graph size manageable and improve overall performance, especially in cases where transformations are frequently added and removed.

Currently, removal of transformations (edges) between is possible, but this does not remove the frames (nodes) themselves. Leading to a growing graph, even when certain frames are no longer connected to others. As the graph grows, recalculating shortest paths via `_recompute_shortest_path` becomes increasingly expensive, contributing to performance degradation with large graphs (noticeable in `remove_transform` when `_recompute_shortest_path` is called).

The remove_frame method addresses this issue by allowing users to remove frames that are no longer involved in any transformations. This reduces the size of the graph, helping to manage complexity.


AlexanderFabisch:
Hi @clungzta ,

great contribution, thank you!

We would need type hints in `_transform_graph_base.pyi` and unit tests. The unit tests should cover all branches and explicitly check that the frames are actually removed.

Let me know if you need help.

pytransform3d/transform_manager/_transform_graph_base.py
```diff
@@ -177,6 +177,53 @@ def remove_transform(self, from_frame, to_frame):
...
+        """
+        if frame not in self.nodes:
+            raise KeyError(f"Frame '{frame}' is not in the graph.")
+
+        # Find the index of the frame to remove
+        frame_index = self.nodes.index(frame)
+
+        # Remove all transformations (edges) associated with the frame
+        transforms_to_remove = []
+        for (from_frame, to_frame) in self.transform_to_ij_index.keys():
```
AlexanderFabisch, line 202: I'd suggest to extract this to a function `_find_connected_transforms(frame)`.

pytransform3d/transform_manager/_transform_graph_base.py
```diff
@@ -177,6 +177,53 @@ def remove_transform(self, from_frame, to_frame):
...
+        Returns
+        -------
+        self : TransformManager
+            This object for chaining.
+        """
+        if frame not in self.nodes:
+            raise KeyError(f"Frame '{frame}' is not in the graph.")
+
+        # Find the index of the frame to remove
+        frame_index = self.nodes.index(frame)
```
AlexanderFabisch, line 198: `frame_index` is only used in line 210. This should be moved down a little bit.

pytransform3d/transform_manager/_transform_graph_base.py
```diff
@@ -177,6 +177,53 @@ def remove_transform(self, from_frame, to_frame):
...
+        # Remove all transformations (edges) associated with the frame
+        transforms_to_remove = []
+        for (from_frame, to_frame) in self.transform_to_ij_index.keys():
+            if from_frame == frame or to_frame == frame:
+                transforms_to_remove.append((from_frame, to_frame))
+        
+        for (from_frame, to_frame) in transforms_to_remove:
+            self.remove_transform(from_frame, to_frame)
+
+        # Remove the frame from the node list
```
AlexanderFabisch, line 209: This is self-explanatory if you know Python, doesn't need a comment.

pytransform3d/test/test_transform_manager.py
```diff
@@ -25,6 +25,38 @@ def test_request_added_transform():
...
+
+    assert tm.has_frame("A")
+    assert tm.has_frame("C")
+    assert tm.has_frame("D")
+
+    # Ensure we cannot retrieve paths involving the removed frame
+    with pytest.raises(KeyError, match="Cannot compute path"):
+        tm.get_transform("A", "C")
+
+    A2D = TM.get_transform("A", "D")
```
AlexanderFabisch, line 57: So far this unit test does not test if some of the transformations are mixed up through index errors. This could be tested by assigning different transformations and checking the transformation here.

clungzta:
Thanks @AlexanderFabisch. I have updated the test based upon your comments and added the `_find_connected_transforms` function. The test of 592adb1 passes in my dev environment.

AlexanderFabisch:
Thanks. I had to fix the unit test a bit in e7c0ea1b10450860cfaf1ba271f106593e68dfc9. Looks good now and was merged to develop. Thanks for the contribution!

PR #301, opened 2024-10-17
Vectorize TemporalTransformManager

JonasHablitzel:
Hello,

i want to add a feature to the libary to support multiple timestamps for the `get_transform_at_time` method for the `TemporalTransformManager` class.

For this i already added a class `NumpyTimeseriesTransformTrajectory`  which is able to take a array as input for the `as_matrix` method.

Regarding the public facing API i wanted to ask you what you would prefer more. For my point of view there are two options:

## Option1
Keep interface the same
```python
manager = TemporalTransformManager()
...
timestamps = np.asarray([10,20,30])
manager .get_transform_at_time("a","b",timestamps )
```

## Option2
add a new method for batch/trajectorie operations.
```python
manager = TemporalTransformManager()
...
timestamps = np.asarray([10,20,30])
manager .get_transform_at_time_batch("a","b",timestamps )
```

I got  **Option1** kinda to work with the following addition.
```python
class TemporalTransformManagerVectorized(TemporalTransformManager):
    def _path_transform(self, path):
        """Convert sequence of node names to rigid transformation."""
        A2B = np.eye(4)
        for from_f, to_f in zip(path[:-1], path[1:]):
            A2B = concat_dynamic(
                A2B,
                self.get_transform(from_f, to_f),
            )
        return A2B
```

## PR Summary

* Add `batch_rotations.norm_axis_angles`
* Add `batch_rotations.axis_angles_from_quaternions`
* Add `trajectories.concat_many_to_many`
* Add `trajectories.concat_dynamic`
* Add `trajectories.screw_parameters_from_dual_quaternions`
* Add `trajectories.dual_quaternions_from_screw_parameters`
* Add `trajectories.dual_quaternions_power`
* Add `trajectories.dual_quaternions_sclerp`
* Add `trajectories.batch_dq_q_conj`
* Vectorize `TimeVaryingTransform.as_matrix` and `TemporalTransformManager.current_time` to enable querying transformations for multiple times at once


AlexanderFabisch:
Hi @JonasHablitzel ,

in my opinion it is not necessary to introduce another subclass of `TimeVaryingTransform` that behaves differently from the others when you actually want to modify the interface of the `TemporalTransformManager`. I would be in favor of modifying the existing `NumpyTimeseriesTransform` that already stores a trajectory. It should be possible to vectorize the current implementation.

I would be a little bit more in favor of `get_transform_at_time("a", "b", timestamps)` vs. `get_transform_at_time_batch("a", "b", timestamps)` as it is a bit more in line with, e.g., the numpy ufuncs.

I would like to involve @kopytjuk who implemented the `TemporalTransformManager` (if you are available).

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -104,9 +103,11 @@ def check_transforms(self):
         return self
 
     def _interpolate_pq_using_sclerp(self, query_time):
+        if isinstance(query_time,float) or isinstance(query_time,float):
```
kopytjuk, line 106: We have to adapt typings [here](https://github.com/dfki-ric/pytransform3d/blob/97909fd124bed1dc560772c4bb08eca9efec958b/pytransform3d/transform_manager/_temporal_transform_manager.pyi#L78) as well since we allow both floats and arrays

pytransform3d/test/test_transform_manager.py
```diff
@@ -504,6 +504,47 @@ def test_numpy_timeseries_transform():
     assert origin_of_A_in_B_y == pytest.approx(-1.28, abs=1e-2)
 
 
+def test_numpy_timeseries_transform_multiple_timestamps_query():
```
kopytjuk, line 507: Since `pytest` is used, consider extending the existing tests using `pytest.mark.parametrize`:

```python
@pytest.mark.paramertrize("query_times", [4.9, np.array([4.9, 4.9])]
def test_numpy_timeseries_transform(query_times):
    # ...
    tm.get_transform_at_time("A", "B", query_times)
```

pytransform3d/test/test_transform_manager.py
```diff
@@ -504,6 +504,47 @@ def test_numpy_timeseries_transform():
...
+
+    tm.add_transform("A", "W", transform_WA)
+    tm.add_transform("B", "W", transform_WB)
+
+    query_time = time_A[0]  # start time
+    A2W_at_start = pt.transform_from_pq(pq_arr_A[0, :])
+    A2W_at_start_2 = tm.get_transform_at_time("A", "W", query_time)
+    assert_array_almost_equal(A2W_at_start, A2W_at_start_2, decimal=2)
+
+    query_times = [4.9,4.9]  # [s]
```
kopytjuk, line 533: This PR adds quite a lot vectorized mathematical operations, which IMHO need a little more unittests than testing with an array of two equal values ...

kopytjuk (commented):
First of all @JonasHablitzel, thank you for your contribution!

I believe making the transform lookup vectorized is a great addition to the code, for many real world projects looping over the timesteps and calling the existing `get_transform_at_time()` is indeed quite slow ...

pytransform3d/trajectories.py
```diff
@@ -109,6 +109,79 @@ def concat_many_to_one(A2Bs, B2C):
...
+        Concatenate one transformation with multiple transformations.
+
+    pytransform3d.transformations.concat :
+        Concatenate two transformations.
+    """
+    return np.einsum("ijk,ikl->ijl", B2C, A2B)
+
+
+def concat_dynamic(A2B, B2C):
+    """Concatenate multiple transformations A2B with transformation B2C.
```
kopytjuk, line 143: To clarify what "dynamic" means:

```suggestion
def concat_dynamic(A2B, B2C):
    """Concatenate multiple transformations A2B with transformation B2C.
    This function accepts different numbers or shapes of the input transformations, e.g. 4x4 or Nx4x4 shape.
```

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+        is represented by [1, 0, 0, 0].
+    """
+    angle = a[:, 3]  # Extract the angle component (N,)
+    norm = np.linalg.norm(a[:, :3], axis=1)  # Compute the norm of the first three elements (axis)
+
+
+    res = np.ones_like(a)
+
+    # Create masks for elements where angle or norm is zero
+    zero_mask = (angle == 0.0) | (norm == 0.0)
```
kopytjuk, line 535: what if a user inputs `angle == 2*np.pi`?

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+
+
+    non_zero_mask = ~zero_mask
+    res[non_zero_mask, :3] = a[non_zero_mask, :3] / norm[non_zero_mask, None]
```
kopytjuk, line 539: A more "numpy" style:

```suggestion
    res[non_zero_mask, :3] = a[non_zero_mask, :3] / norm[non_zero_mask, :]  
```

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+            (translation[outer_else_mask] - distance[:, None] * s_axis[outer_else_mask])
+            / np.tan(0.5 * thetas[outer_else_mask])[:, None]
+        )
+
+        qs[outer_else_mask] = np.cross(s_axis[outer_else_mask], moment)
+        hs[outer_else_mask] = distance / thetas[outer_else_mask]
+
+    return qs, s_axis, hs, thetas
+
+def dual_quaternions_from_screw_parameters(qs, s_axis, hs, thetas):
```
kopytjuk, line 670: Since this function is non-trivial when it comes to mathematical operations, would you please add at least a unittest so that a next developer wants to optimize/change/accelerate could check her work?

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+    w_clamped = np.clip(qs[:, 0], -1.0, 1.0)  # Clamp the scalar part of the quaternion
+    angle = 2.0 * np.arccos(w_clamped)
+
+    # Stack the axis and the angle together and normalize the axis-angle representation
+    result[non_zero_mask] = norm_axis_angles(np.hstack((axis[non_zero_mask], angle[non_zero_mask, None])))
+
+    return result
+
+
+def screw_parameters_from_dual_quaternions(dqs):
```
kopytjuk, line 594: Since this function is non-trivial when it comes to mathematical operations, would you please add at least a unittest so that a next developer wants to optimize/change/accelerate could check her work?

kopytjuk:
The more I look deeper into this contribution, the more I like it. 

I think a good solution would be to:

1. extend the `TemporalTransformManager` to have an array of timestamps as `self._current_time(s)` (instead of a scalar time) 
2. with that, the `TimeVaryingTransform.as_matrix()` should accept time arrays as well. @JonasHablitzel already kind of addressed that in the `NumpyTimeseriesTransform` ✅ 
3. Having that, we can use the proposed adaptation of `_path_transform` with the `concat_dynamic` directly in the `TransformGraphBase`.

pytransform3d/transform_manager/_temporal_transform_manager.py
AlexanderFabisch: Also this line does not really make sense, right? Checks the same thing twice. It is also not very general. What if an int is passed? Or float32?

pytransform3d/test/test_transform_manager.py
AlexanderFabisch: Starting with the nitpicking here: please also check for PEP8 with either pep8 or flake8.

AlexanderFabisch:
> What do you think @AlexanderFabisch on the plan?

Sounds good to me.

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+
+    Returns
+    -------
+    a : array, shape (...,4)
+        Axis of rotation and rotation angle: (x, y, z, angle). The length
+        of the axis vector is 1 and the angle is in [0, pi). No rotation
+        is represented by [1, 0, 0, 0].
+    """
+    angle = a[:, 3]  # Extract the angle component (N,)
+    norm = np.linalg.norm(a[:, :3], axis=1)  # Compute the norm of the first three elements (axis)
```
AlexanderFabisch, line 529: You can use `a[..., :3]` to allow an arbitrary number of dimensions.

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+        Axis of rotation and rotation angle: (x, y, z, angle)
+
+    Returns
+    -------
+    a : array, shape (...,4)
+        Axis of rotation and rotation angle: (x, y, z, angle). The length
+        of the axis vector is 1 and the angle is in [0, pi). No rotation
+        is represented by [1, 0, 0, 0].
+    """
+    angle = a[:, 3]  # Extract the angle component (N,)
```
AlexanderFabisch, line 528: The variable name explains it. The comment is not necessary. I would also call the variable `angles`.

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+    """Normalize axis-angle representation.
+
+    Parameters
+    ----------
+    a : array-like, shape (...,4)
+        Axis of rotation and rotation angle: (x, y, z, angle)
+
+    Returns
+    -------
+    a : array, shape (...,4)
```
AlexanderFabisch, line 523: PEP8 probably would detect this: please add spaces after commas.

pytransform3d/trajectories.py
```diff
@@ -109,6 +109,79 @@ def concat_many_to_one(A2Bs, B2C):
...
+        Concatenate two transformations.
+    """
+    if B2C.ndim == 2 and A2B.ndim == 2:
+        return B2C.dot(A2B)
+    if B2C.ndim == 2 and A2B.ndim == 3:
+        return concat_many_to_one(A2B, B2C)
+    if B2C.ndim == 3 and A2B.ndim == 2:
+        return concat_one_to_many(A2B, B2C)
+    if B2C.ndim == 3 and A2B.ndim == 3:
+        return concat_many_to_many(A2B, B2C)
```
AlexanderFabisch, line 182: What if none of the conditions is true?

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+    q : array-like, shape (...,4)
+        Unit quaternion to represent rotation: (w, x, y, z)
+
+    Returns
+    -------
+    a : array, shape (...,4)
+        Axis of rotation and rotation angle: (x, y, z, angle). The angle is
+        constrained to [0, pi) so that the mapping is unique.
+    """
+    ps = qs[:, 1:]  # Extract the vector part of the quaternion
```
AlexanderFabisch, line 570: Again `qs[..., 1:]` would be more general and according to the docstring.

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+    p_norm = np.linalg.norm(ps, axis=1)  # Compute the norm of the vector part
```
AlexanderFabisch, line 571: `axis=-1` is the general version

AlexanderFabisch: I think we also don't need a comment for this line. It is clear that the function `norm` computes a norm.

pytransform3d/trajectories.py
AlexanderFabisch: btw. if you need a name that explains what the variable contains, consider changing the name of the variable, e.g., `quaternion_vector_part`

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+
+    # Create a mask for quaternions where p_norm is small
```
AlexanderFabisch, line 573: Comment says the same as variable name

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+    small_p_norm_mask = p_norm < np.finfo(float).eps
+
+    # Initialize the output with default values for small p_norm cases
+    result = np.zeros_like(qs)
+    result[small_p_norm_mask] = np.array([1.0, 0.0, 0.0, 0.0])
+
+    # For non-zero norms, calculate axis, clamped w, and angle
+    non_zero_mask = ~small_p_norm_mask
+    axis = np.zeros_like(ps)
+    axis[non_zero_mask] = ps[non_zero_mask] / p_norm[non_zero_mask, None]
```
AlexanderFabisch, line 583: Again, I would use `np.newaxis`, not `None`. This is more readable.

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+    angle = 2.0 * np.arccos(w_clamped)
+
+    # Stack the axis and the angle together and normalize the axis-angle representation
+    result[non_zero_mask] = norm_axis_angles(np.hstack((axis[non_zero_mask], angle[non_zero_mask, None])))
+
+    return result
+
+
+def screw_parameters_from_dual_quaternions(dqs):
+    """Compute screw parameters from dual quaternion.s
```
AlexanderFabisch, line 595: typo

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+
+    Parameters
+    ----------
+    dq : array-like, shape (…,8)
```
AlexanderFabisch, line 599: Please replace `…` by `...`

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+        of the screw axis. Infinite pitch indicates pure translation.
+
+    thetas : array, shape (…,)
+        Parameter of the transformation: theta is the angle of rotation
+        and h * theta the translation.
+    """
+    reals = dqs[:, :4]
+    duals = dqs[:, 4:]
+
+    # Vectorized axis angle from quaternion
```
AlexanderFabisch, line 622: The comment says the same as the function name

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+    a = axis_angles_from_quaternions(reals)  # Should handle vectorized input
```
AlexanderFabisch, line 623: comment is not necessary

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+
+    hs : array, shape (…,)
+        Pitch of the screw. The pitch is the ratio of translation and rotation
+        of the screw axis. Infinite pitch indicates pure translation.
+
+    thetas : array, shape (…,)
+        Parameter of the transformation: theta is the angle of rotation
+        and h * theta the translation.
+    """
+    reals = dqs[:, :4]
```
AlexanderFabisch, line 619: replace `:` by `...`

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+    if np.any(h_is_inf_mask):
+        ds[h_is_inf_mask] = thetas[h_is_inf_mask]
+        mod_thetas[h_is_inf_mask] = 0
+    
+    if np.any(h_is_not_inf_mask):
+        ds[h_is_not_inf_mask] = hs[h_is_not_inf_mask] * thetas[h_is_not_inf_mask]
+
+    moments = np.cross(qs, s_axis)
+    half_distances = 0.5 * ds
+    sin_half_angles = np.sin(0.5 * mod_thetas)
```
AlexanderFabisch, line 686: `0.5 * mod_thetas` could be stored in a temporary variable. Since this function could be applied to large arrays, we should do this.

pytransform3d/trajectories.py
```diff
@@ -436,6 +509,201 @@ def pqs_from_dual_quaternions(dqs):
...
+    cos_half_angles = np.cos(0.5 * mod_thetas)
+
+    real_w = cos_half_angles
+    real_vec = sin_half_angles[..., None] * s_axis
```
AlexanderFabisch, line 690: `None` -> `np.newaxis`

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -104,9 +103,11 @@ def check_transforms(self):
...
+            query_time_arr = np.array([query_time])
```
AlexanderFabisch, line 107: We can just use `np.atleast_1d` for this. Then we don't need to branch.

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -104,9 +103,11 @@ def check_transforms(self):
...
+
         # identify the index of the preceding sample
-        idx_timestep_earlier_wrt_query_time = np.argmax(
-            self.time >= query_time) - 1
+        idx_timestep_earlier_wrt_query_time = np.searchsorted(self.time, query_time_arr, side='right') - 1
```
AlexanderFabisch, line 110: Can this result in `-1`?

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -115,18 +116,20 @@ def _interpolate_pq_using_sclerp(self, query_time):
...
+        dq_next = dual_quaternions_from_pqs(pq_next)
 
         # since sclerp works with relative (0-1) positions
         rel_delta_t = (query_time - t_prev) / (t_next - t_prev)
-        dq_interpolated = dual_quaternion_sclerp(dq_prev, dq_next, rel_delta_t)
-
-        return pq_from_dual_quaternion(dq_interpolated)
+        dqs_interpolated = dual_quaternions_sclerp(dq_prev, dq_next, rel_delta_t)
+        res = pqs_from_dual_quaternions(dqs_interpolated)
+        if res.shape[0] == 1:
```
AlexanderFabisch, line 130: I would prefer a solution without a branch.

AlexanderFabisch:
I had the chance to review the PR now. I think the solution will be very elegant in the end. I really like it!

AlexanderFabisch:
Also please have a look at

* https://github.com/dfki-ric/pytransform3d/wiki#tools
* https://github.com/dfki-ric/pytransform3d/wiki#checklist-for-new-features

pytransform3d/trajectories.py
JonasHablitzel: i would go with raising a ValueError for now, would this also the solution you prefer here?

AlexanderFabisch: I don't have a better idea. Yes.

JonasHablitzel:
Hello,
i added the suggested changes into the code and added some docstrings and some unittests.
Regearding the `TemporalTransformManager` i needed to redefine the `get_transform` as there was a problem when the `invert_transform` got called with a array off size (Nx4x4) so i had to use the `invert_transforms`. The external API stayed the same, so when a user passes in a float he will get back a (4x4) when he passes a array of times he will get a (Nx4x4) array.


- [x] are there tests for every branch and special case?
      - not all special cases i oriented myself on the existing tests for the trajetories
- [x] does it have docstrings?
- [x] does it have type hints in the corresponding .pyi file?
- [x] is it included in the API documentation
- [x] run flake8, pylint, and mypy
- [x] should it be part of the user documentation?
- [x] should it be included in any example script?
      -  i included the additional functions and also a hin in the timneseries section that we can also pass in arrays

pytransform3d/test/test_trajectories.py
```diff
@@ -296,3 +299,71 @@ def test_mirror_screw_axis():
...
+
+
+def test_dual_quaternions_from_screw_parameters():
+    q_0 = np.zeros(3)
+    s_axis_0 = np.array([1, 0, 0])
+    h_0 = np.inf
+    theta_0 = 0.0
+
+    q_1 = np.zeros(3)
+    s_axis_1 = np.array([0.55297409, 0.57701644, 0.6010588 ])#pr.norm_vector(np.array([2.3, 2.4, 2.5]))
```
AlexanderFabisch, line 331: I would either use the function call or remove the comment.

pytransform3d/test/test_transform_manager.py
```diff
@@ -489,6 +488,13 @@ def test_numpy_timeseries_transform():
     A2W_at_start = pt.transform_from_pq(pq_arr_A[0, :])
     A2W_at_start_2 = tm.get_transform_at_time("A", "W", query_time)
     assert_array_almost_equal(A2W_at_start, A2W_at_start_2, decimal=2)
+    assert A2W_at_start_2.ndim == 2
+
+    query_times = [time_A[0],time_A[0]] # start times
```
AlexanderFabisch, line 493: PEP8

pytransform3d/test/test_transform_manager.py
```diff
@@ -521,3 +527,38 @@ def test_numpy_timeseries_transform_wrong_input_shapes():
...
+    )
+    transform_WB = NumpyTimeseriesTransform(time_B, pq_arr_B)
+
+    tm = TemporalTransformManager()
+
+    tm.add_transform("A", "W", transform_WA)
+    tm.add_transform("B", "W", transform_WB)
+
+
+    query_times = np.array([4.9,5.2])  # [s]
```
AlexanderFabisch, line 553: PEP8

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -24,13 +31,13 @@ def as_matrix(self, query_time):
...
-        query_time : float
+        query_time : Union[float,array-like shape (...)]
             Query time
 
         Returns
         -------
-        A2B_t : array, shape (4, 4)
-            Homogeneous transformation matrix at given time.
+        A2B_t : array, shape (4, 4) or (..., 4, 4)
+            Homogeneous transformation matrix at given time. or times
```
AlexanderFabisch, line 40: Placement of "."

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -44,6 +51,7 @@ def check_transforms(self):
         """
```
AlexanderFabisch, line 53: PEP8

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -96,37 +105,61 @@ def __init__(self, time, pqs):
...
+
+        Returns
+        -------
+        A2B_t : array, shape (4, 4) or (..., 4, 4)
+            Homogeneous transformation matrix at given time. . or times
+        """
         pq = self._interpolate_pq_using_sclerp(query_time)
-        return transform_from_pq(pq)
+        transforms = transforms_from_pqs(pq)
+        transforms = np.squeeze(transforms)  # to keep the external API
```
AlexanderFabisch, line 122: What if the shape of `query_time` is `(1, 1)`?

I would use something like

```python
query_time = np.array(query_time)
transforms.reshape(*(list(query_time.shape) + [4, 4]))
```

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -96,37 +105,61 @@ def __init__(self, time, pqs):
...
+        idxs_timestep_earlier_wrt_query_time = np.searchsorted(
+            self.time, query_time_arr, side='right'
+        ) - 1
 
-        # deal with first timestamp
-        idx_timestep_earlier_wrt_query_time = max(
-            idx_timestep_earlier_wrt_query_time, 0)
+
+        # deal with first and last timestamp
+        min_index = 0
```
AlexanderFabisch, line 139: Does the unit test check if edge cases are handled correctly?

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -201,6 +235,33 @@ def get_transform_at_time(self, from_frame, to_frame, time):
         self.current_time = previous_time
         return A2B
 
+    def get_transform(self, from_frame, to_frame):
+        # overwrite get_transform to be able to work with arrys trajectories
```
AlexanderFabisch, line 239: Typo "arrys"

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -201,6 +235,33 @@ def get_transform_at_time(self, from_frame, to_frame, time):
...
+        if self.check:
```
AlexanderFabisch, line 240: Does the unit test cover all the branches?

pytransform3d/transform_manager/_temporal_transform_manager.pyi
```diff
@@ -10,7 +10,7 @@ from ._transform_graph_base import TransformGraphBase
 class TimeVaryingTransform(abc.ABC):
 
     @abc.abstractmethod
-    def as_matrix(self, query_time: float) -> np.ndarray: ...
+    def as_matrix(self, query_time: Union[float,np.ndarray]) -> np.ndarray: ...
```
AlexanderFabisch, line 13: PEP8

pytransform3d/transform_manager/_temporal_transform_manager.pyi
```diff
@@ -21,7 +21,7 @@ class StaticTransform(TimeVaryingTransform):
 
     def __init__(self, A2B: npt.ArrayLike): ...
 
-    def as_matrix(self, query_time: float) -> np.ndarray: ...
+    def as_matrix(self, query_time: Union[float,np.ndarray]) -> np.ndarray: ...
```
AlexanderFabisch, line 24: PEP8

pytransform3d/transform_manager/_temporal_transform_manager.pyi
```diff
@@ -75,4 +75,4 @@ class TemporalTransformManager(TransformGraphBase):
 
     def get_transform_at_time(self, from_frame: Hashable,
                               to_frame: Hashable,
-                              query_time: float) -> np.ndarray: ...
+                              query_time: Union[float,np.ndarray]) -> np.ndarray: ...
```
AlexanderFabisch, line 78: PEP8

pytransform3d/test/test_transform_manager.py
```diff
@@ -521,3 +527,38 @@ def test_numpy_timeseries_transform_wrong_input_shapes():
         time = np.arange(10)
         pqs = np.random.randn(n_steps, 8).flatten()
         NumpyTimeseriesTransform(time, pqs)
+
```
AlexanderFabisch, line 530: PEP8

pytransform3d/test/test_transform_manager.py
AlexanderFabisch: Sounds like a good idea. Can we do this?

pytransform3d/trajectories.py
AlexanderFabisch: Comment should be removed

pytransform3d/trajectories.py
```diff
@@ -436,6 +515,322 @@ def pqs_from_dual_quaternions(dqs):
...
+
+    res = np.ones_like(a)
+
+    # Create masks for elements where angle or norm is zero
+    zero_mask = (angles == 0.0) | (norm == 0.0)
+
+
+    non_zero_mask = ~zero_mask
+    res[non_zero_mask, :3] = (
+        a[non_zero_mask, :3] / norm[non_zero_mask, np.newaxis]
```
AlexanderFabisch, line 546: see @kopytjuk 's comment: is np.newaxis really required?

pytransform3d/trajectories.py
```diff
@@ -436,6 +515,322 @@ def pqs_from_dual_quaternions(dqs):
...
+
+    Returns
+    -------
+    a : array, shape (..., 4)
+        Axis of rotation and rotation angle: (x, y, z, angle). The length
+        of the axis vector is 1 and the angle is in [0, pi). No rotation
+        is represented by [1, 0, 0, 0].
+    """
+    angles = a[..., 3]
+    norm = np.linalg.norm(a[..., :3], axis=1)
```
AlexanderFabisch, line 535: should be `axis=-1` now

pytransform3d/trajectories.py
```diff
@@ -436,6 +515,322 @@ def pqs_from_dual_quaternions(dqs):
...
+
+    return result
+
+
+def screw_parameters_from_dual_quaternions(dqs):
+    """Compute screw parameters from dual quaternions.
+
+    Parameters
+    ----------
+    dqs : array-like, shape (...,8)
```
AlexanderFabisch, line 617: space between , and 8

pytransform3d/trajectories.py
```diff
@@ -436,6 +515,322 @@ def pqs_from_dual_quaternions(dqs):
...
+        Unit dual quaternion to represent transform:
+        (pw, px, py, pz, qw, qx, qy, qz)
+
+    Returns
+    -------
+    qs : array, shape (...,3)
```
AlexanderFabisch, line 623: space

pytransform3d/trajectories.py
```diff
@@ -436,6 +515,322 @@ def pqs_from_dual_quaternions(dqs):
...
+        Vector to a point on the screw axis
+
+    s_axiss : array, shape (...,3)
```
AlexanderFabisch, line 626: space

pytransform3d/trajectories.py
```diff
@@ -436,6 +515,322 @@ def pqs_from_dual_quaternions(dqs):
...
+        Compute screw parameters from dual quaternion.
+    """
+    reals = dqs[..., :4]
+    duals = dqs[..., 4:]
+
+    a = axis_angles_from_quaternions(reals)
+    s_axis = a[..., :3]
+    thetas = a[..., 3]
+
+    # Vectorized quaternion concatenation and translation computation
```
AlexanderFabisch, line 649: Comment does not add a lot of information

pytransform3d/trajectories.py
```diff
@@ -436,6 +515,322 @@ def pqs_from_dual_quaternions(dqs):
...
+        mod_thetas[h_is_inf_mask] = 0
+
+    if np.any(h_is_not_inf_mask):
+        ds[h_is_not_inf_mask] = hs[h_is_not_inf_mask] * \
+            thetas[h_is_not_inf_mask]
+
+    moments = np.cross(qs, s_axis)
+    half_distances = 0.5 * ds
+    half_thetas = 0.5 * mod_thetas
+    sin_half_angles = np.sin(0.5 * half_thetas)
```
AlexanderFabisch, line 749: This should now be an error. We compute `sin(0.5 * 0.5 * mod_thetas)`, it was `sin(0.5 * mod_thetas)` before

AlexanderFabisch: So it seems like there is also no unit test that checks this function properly.

pytransform3d/trajectories.py
```diff
@@ -487,6 +882,37 @@ def batch_dq_conj(dqs):
...
+def batch_dq_q_conj(dqs):
+    """Quaternion conjugate of dual quaternions.
+
+    For unit dual quaternions that represent transformations,
+    this function is equivalent to the inverse of the
+    corresponding transformation matrix.
+
+    Parameters
+    ----------
+    dqs : array-like, shape (8,)
```
AlexanderFabisch, line 894: shape (..., 8)

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -24,13 +31,13 @@ def as_matrix(self, query_time):
 
         Parameters
         ----------
-        query_time : float
+        query_time : Union[float,array-like shape (...)]
```
AlexanderFabisch, line 34: space after float,

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -52,6 +60,7 @@ class StaticTransform(TimeVaryingTransform):
     A2B : array-like, shape (4, 4)
         Homogeneous transformation matrix.
     """
+
```
AlexanderFabisch, line 63: PEP8

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -96,37 +105,61 @@ def __init__(self, time, pqs):
...
 
         # since sclerp works with relative (0-1) positions
         rel_delta_t = (query_time - t_prev) / (t_next - t_prev)
-        dq_interpolated = dual_quaternion_sclerp(dq_prev, dq_next, rel_delta_t)
-
-        return pq_from_dual_quaternion(dq_interpolated)
+        dqs_interpolated = dual_quaternions_sclerp(
+            dq_prev, dq_next, rel_delta_t)
+        res = pqs_from_dual_quaternions(dqs_interpolated)
+        return res
```
AlexanderFabisch, line 162: If we pass a float query_time, this should return an array of shape (4, 4).

My solution to this would be to put as a first line of the function `query_shape = np.asarray(query_time).shape` and reshape with `res.reshape(*(list(query_shape) + [4, 4]))`.

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -178,8 +212,8 @@ def get_transform_at_time(self, from_frame, to_frame, time):
         to_frame : Hashable
             Name of the frame in which the transformation is defined
 
-        time : float
-            Time at which we request the transformation.
+        time : Union[float,array-like shape (...)]
```
AlexanderFabisch, line 215: space after float,

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -178,8 +212,8 @@ def get_transform_at_time(self, from_frame, to_frame, time):
...
+            Time or Times at which we request the transformation.
```
AlexanderFabisch, line 216: lower case: `times`

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -201,6 +235,33 @@ def get_transform_at_time(self, from_frame, to_frame, time):
         self.current_time = previous_time
         return A2B
 
+    def get_transform(self, from_frame, to_frame):
```
AlexanderFabisch, line 238: function should have a docstring

pytransform3d/transform_manager/_temporal_transform_manager.pyi
```diff
@@ -32,27 +32,27 @@ class NumpyTimeseriesTransform(TimeVaryingTransform):
 
     def __init__(self, time: npt.ArrayLike, pqs: npt.ArrayLike): ...
 
-    def as_matrix(self, query_time: float) -> np.ndarray: ...
+    def as_matrix(self, query_time: Union[float,np.ndarray]) -> np.ndarray: ...
```
AlexanderFabisch, line 35: space after float,

pytransform3d/transform_manager/_temporal_transform_manager.pyi
```diff
@@ -32,27 +32,27 @@ class NumpyTimeseriesTransform(TimeVaryingTransform):
...
 
     def check_transforms(self) -> "NumpyTimeseriesTransform": ...
 
-    def _interpolate_pq_using_sclerp(self, query_time) -> np.ndarray: ...
+    def _interpolate_pq_using_sclerp(self, query_time: Union[float,np.ndarray]) -> np.ndarray: ...
```
AlexanderFabisch, line 39: space after float,

AlexanderFabisch:
@JonasHablitzel I went through the PR again.

Since a new version of pytransform3d was released recently, I'd suggest to rebase this PR on the latest development branch. There are some changes in how the API documentation is sorted.

Before we finalize the PR, we also have to make sure that all new functions are properly tested. If you give me editing rights on your fork, I can help you with that.

AlexanderFabisch:
I added two test cases for the trajectories module. The temporal transform manager does not have full coverage yet. However, before we finish this, please rebase this branch on develop.

pytransform3d/trajectories.py
JonasHablitzel: yes as norm is a 1D array and a has size (N, 3) we need to transform norm to a (N, 1) to get it correctly broadcastet and we want to perform element wise division here.

pytransform3d/trajectories.py
```diff
@@ -436,6 +515,324 @@ def pqs_from_dual_quaternions(dqs):
...
+    --------
+    pytransform3d.transformations.dual_quaternion_power :
+        Compute power of unit dual quaternion with respect to scalar.
+    """
+    q, s_axis, h, theta = screw_parameters_from_dual_quaternions(dqs)
+    return dual_quaternions_from_screw_parameters(q, s_axis, h, theta * ts)
+
+
+def dual_quaternions_sclerp(starts, ends, ts):
+    """Screw linear interpolation (ScLERP) for arry of dual quaternions.
```
AlexanderFabisch, line 795: arry -> array

JonasHablitzel:
sorry for the many pep8 errors btw, i mostly a black user so i am not so used to do the formatting manually.

pytransform3d/trajectories.py
```diff
@@ -436,6 +515,324 @@ def pqs_from_dual_quaternions(dqs):
...
+        Pitch of the screw. The pitch is the ratio of translation and rotation
+        of the screw axis. Infinite pitch indicates pure translation.
+
+    thetas : array-like, shape (...,)
+        Parameter of the transformation: theta is the angle of rotation
+        and h * theta the translation.
+
+    Returns
+    -------
+    dqs : array, shape (...,8)
```
AlexanderFabisch: space in `(..., 8)`

pytransform3d/trajectories.py
```diff
@@ -436,6 +515,324 @@ def pqs_from_dual_quaternions(dqs):
...
+    dqs : array-like, shape (..., 8)
+        Unit dual quaternions to represent transform:
+        (pw, px, py, pz, qw, qx, qy, qz)
+
+    t : array-like, shape (...)
+        Exponent
+
+    Returns
+    -------
+    dq_ts : array, shape (...,8)
```
AlexanderFabisch: space

AlexanderFabisch:
@JonasHablitzel It seems like you did a merge instead of a rebase with this branch? At least this is my explanation why GitHub messes up the diff to develop. Could you try to rebase? I don't want to mess around with your fork. The procedure should be roughly:

```bash
git checkout develop
git remote add original https://github.com/dfki-ric/pytransform3d.git
git pull original develop  # where original points to this repository, not your fork
git checkout feat/trajectot_temporal_transform  # switch back to your branch
git rebase develop
```

Then you might have to resolve several conflicts, since the merged commits are now twice in your commit history: from the merge and from the rebase.

kopytjuk:
If you guys need an additional reviewer (who is not an active developer here) let me know :)

AlexanderFabisch:
> If you guys need an additional reviewer (who is not an active developer here) let me know :)

Thanks a lot. There are still some things to test before we can do the final review though:

* [ ] concat_many_to_many
* [ ] screw_parameters_from_dual_quaternions with 1D and 3D input
* [ ] batch_dq_q_conj
* [ ] dual_quaternions_power
* [ ] dual_quaternions_from_screw_parameters with 1D and 3D input
* [ ] full coverage for TemporalTransformManager

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -95,38 +102,74 @@ def __init__(self, time, pqs):
...
     def _interpolate_pq_using_sclerp(self, query_time):
-        # identify the index of the preceding sample
-        idx_timestep_earlier_wrt_query_time = np.argmax(
-            self.time >= query_time) - 1
+        query_time_arr = np.atleast_1d(query_time)
 
-        # deal with first timestamp
-        idx_timestep_earlier_wrt_query_time = max(
-            idx_timestep_earlier_wrt_query_time, 0)
+        # identify the index of the preceding sample
```
AlexanderFabisch, line 133: The handling of edge cases is way more complicated now. I think it is better now, but I'd like to hear your opinion (and need a review because the logic is considerably more complicated) @JonasHablitzel  @kopytjuk .

When the time is beyond the limits of the time array of this TimeVaryingTransform, I clip it to [min, max] and do not interpolate, I just return the first or last element. (Maybe the logic can be simplified a bit still.) Do you think the behavior makes sense. I am not entirely sure how these cases were handled before. I think an interpolation between the two first or last elements was done. Is that correct?

JonasHablitzel: I like the more complicated approach :) 
jup at least in my implementation this was  the case.

I also thought of to maybe expose a configuration flag in the `TimeVaryingTransform` if a user wants to have "extrapolation" or not. Because i think this can differ from usecase to usecase.

the flag could be:
```python
allow_extrapolation:bool=True
```

- if allow_extrapolation is enabled the logic would be the same as @AlexanderFabisch  suggested
- if its disabeld we would raise a ValueError if a tranform with a time larger or smaller then the [min,max] time is requested

what is your opinion on that? @AlexanderFabisch @kopytjuk

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -95,38 +102,74 @@ def __init__(self, time, pqs):
...
+        idxs_timestep_earlier_wrt_query_time[
+            after_end] = idxs_timestep_later_wrt_query_time[after_end]
 
         # dual quaternion from preceding sample
-        t_prev = self.time[idx_timestep_earlier_wrt_query_time]
-        pq_prev = self._pqs[idx_timestep_earlier_wrt_query_time, :]
-        dq_prev = dual_quaternion_from_pq(pq_prev)
+        t_prev = self.time[idxs_timestep_earlier_wrt_query_time]
+        pq_prev = self._pqs[idxs_timestep_earlier_wrt_query_time]
+        dq_prev = dual_quaternions_from_pqs(pq_prev)
```
kopytjuk, line 155: What do you think about more "array-like" names, e.g.
```suggestion
        times_prev = self.time[idxs_timestep_earlier_wrt_query_time]
        pqs_prev = self._pqs[idxs_timestep_earlier_wrt_query_time]
        dqs_prev = dual_quaternions_from_pqs(pq_prev)
```

Same for `_next` dual quaternions ...

pytransform3d/transform_manager/_temporal_transform_manager.py
kopytjuk: For now (to finalize this PR) I would simply raise a ValueError if any of the elements of`query_time` array is outside of the provided transform samples. This would allow us to think about a nice control flag (e.g. as kwarg) in one of the next PRs. With that we can keep the existing logic and rewrite later if we want.

In any case I would avoid any automatisms, like automatic clipping. In my experience people just pass things, it does not crash and they debug for a very long time in their application code why something is wrong ... after hours/days they recognize that their timestamp might be the issue. Often two transforms, be it pq or homogenous matrices are hard to distinguish if they are different ...

AlexanderFabisch:
I'm done with the unit tests. From my perspective we only need to do a final review of this PR and then we can merge it.

pytransform3d/transform_manager/_temporal_transform_manager.py
```diff
@@ -95,38 +107,87 @@ def __init__(self, time, pqs):
         if self._pqs.shape[1] != 7:
             raise ValueError("`pqs` matrix shall have 7 columns.")
 
+        self._min_time = min(self.time)
```
JonasHablitzel, line 110: is there a reason you used the python inbuilts? or could we switch to np.min and np.max respectivly to have better performance?

JonasHablitzel: same further below with the 
```python
elif any(out_of_range):
```

could be changed to a
```python
elif np.any(out_of_range):
```

AlexanderFabisch: In my experience the python builtins are faster as they have less overhead. Although we would have to test this for relevant array sizes.

JonasHablitzel: What are the relevant array sizes?
- lower end, just one transformation at one time (array size 1)
- upper end, Lidar-Pointcloud with time field, tranformation for every point (50_000 Points in a cloud)

i made a small microbenchmark with:
- AMD Ryzen 7 3700X
- Python 3.11.10
- numpy 2.1.3
- tested array sizes:  [5, 10, 50, 100, 500, 1000, 10_000, 100_000]

Verdict:
- inbuild any is always faster then numpy.any
- max/min , numpy is faster for arrays larger then `50`
- for an array of `10000`numpy is `150`-times faster

as the penalty of using numpy for small arrays is only minor i would suggest using `np.min/np.max` instead of the inbuilt ones and leave the inbuilt any. I know that these microbenchmarks can always be kinda missleading/error prone. But i think that the factor 150 for `10000`  is significant enough to make this decision. I attached the benchmark code as well

 
[micro_benchmark.py.txt](https://github.com/user-attachments/files/17871756/micro_benchmark.py.txt)

AlexanderFabisch: Thanks for benchmarking. However, I would expect the overhead of using builtin min/max to be negligible when we have to compute ScLERP for more than 50 transformations. Could you test that? If that is the case, I think it makes more sense to optimize for small numbers of query times as I expect the case of querying 1 or a few times to be much more common.

pytransform3d/test/test_transform_manager.py
```diff
@@ -422,6 +422,60 @@ def test_temporal_transform():
...
+    assert tm.i == [tm.nodes.index("A"), tm.nodes.index("D")]
+    assert tm.j == [tm.nodes.index("D"), tm.nodes.index("E")]
+
+    with pytest.raises(KeyError, match="Unknown frame"):
+        tm.get_transform("A", "B")
+    with pytest.raises(KeyError, match="Unknown frame"):
+        tm.get_transform("B", "C")
+
+    assert tm.has_frame("A")
+    assert tm.has_frame("C")
```
kopytjuk, line 465: Just an API question, do we want `C` to exist after we removed `B` and lost the connection?

pytransform3d/test/test_transform_manager.py
kopytjuk, line 1: Since you are touching this file anyway, can you please add a discriptive comment:

```python
# transformation between W(orld) and A
transform_WA = NumpyTimeseriesTransform(time_A, pq_arr_A)
```

Because when I wrote the test, I did not explain what W means

pytransform3d/test/test_transform_manager.py
```diff
@@ -485,10 +539,24 @@ def test_numpy_timeseries_transform():
...
+    assert A2W_at_start_2.ndim == 2
+
+    query_times = [time_A[0], time_A[0]]  # Start times
+    A2Ws_at_start_2 = tm.get_transform_at_time("A", "W", query_times)
+    assert_array_almost_equal(A2W_at_start, A2Ws_at_start_2[0], decimal=2)
+    assert_array_almost_equal(A2W_at_start, A2Ws_at_start_2[1], decimal=2)
+    assert A2Ws_at_start_2.ndim == 3
+    assert A2W_at_start_2.ndim == 2
+
+    query_times = [time_A[0], time_A[0]]  # start times
```
kopytjuk, line 555: Imho no need to redefine them ...

pytransform3d/test/test_transform_manager.py
```diff
@@ -485,10 +539,24 @@ def test_numpy_timeseries_transform():
...
     A2W_at_start_2 = tm.get_transform_at_time("A", "W", query_time)
     assert_array_almost_equal(A2W_at_start, A2W_at_start_2, decimal=2)
+    assert A2W_at_start_2.ndim == 2
+
+    query_times = [time_A[0], time_A[0]]  # Start times
+    A2Ws_at_start_2 = tm.get_transform_at_time("A", "W", query_times)
+    assert_array_almost_equal(A2W_at_start, A2Ws_at_start_2[0], decimal=2)
+    assert_array_almost_equal(A2W_at_start, A2Ws_at_start_2[1], decimal=2)
+    assert A2Ws_at_start_2.ndim == 3
+    assert A2W_at_start_2.ndim == 2
```
kopytjuk, line 553: Since it is checked above already:

```suggestion
```

pytransform3d/test/test_transform_manager.py
```diff
@@ -521,3 +583,106 @@ def test_numpy_timeseries_transform_wrong_input_shapes():
...
+    )
+    B2world = NumpyTimeseriesTransform(time_B, pq_arr_B, time_clipping=True)
+
+    tm = TemporalTransformManager()
+    tm.add_transform("A", "W", A2world)
+    tm.add_transform("B", "W", B2world)
+
+    assert min(time_A) == 0.0
+    assert min(time_B) == 0.1
+    A2B_at_start_time, A2B_before_start_time = tm.get_transform_at_time(
```
AlexanderFabisch, line 667: @JonasHablitzel This is a great example of how readable the new vectorized version of the interface is. :)

pytransform3d/batch_rotations.py
```diff
@@ -38,6 +38,52 @@ def norm_vectors(V, out=None):
...
+        is represented by [1, 0, 0, 0].
+    """
+    a = np.asarray(a)
+
+    # Handle the case of only one axis-angle instance
+    only_one = a.ndim == 1
+    a = np.atleast_2d(a)
+
+    angles = a[..., 3]
+    norm = np.linalg.norm(a[..., :3], axis=-1)
```
kopytjuk, line 63: You probably wanted to use the variable:

```suggestion
    norm = np.linalg.norm(angles, axis=-1)
```

pytransform3d/batch_rotations.py
```diff
@@ -38,6 +38,52 @@ def norm_vectors(V, out=None):
...
+
+    zero_mask = (angles == 0.0) | (norm == 0.0)
```
kopytjuk, line 65: Maybe a more suitable name (but only my subjective opinion):

```suggestion
    no_rotation_mask = (angles == 0.0) | (norm == 0.0)
```

pytransform3d/batch_rotations.py
```diff
@@ -38,6 +38,52 @@ def norm_vectors(V, out=None):
...
+    non_zero_mask = ~zero_mask
+
+    res = np.empty_like(a)
```
kopytjuk, line 68: About the usage of `empty_like` which is callen on the input argument - are we sure that we want to use it without prior casting to float?

e.g. when `a` is an integer array, assignments `res[non_zero_mask, :3] = ...` are casted to int and may get output wrong results ...

However, I just tested it, and we luckily do have have an unintended "safeguard":

```python
a = np.array([[1, 2, 3, 2], [4, 5, 6, 3]])
# a = a.astype(float)  # works if uncommented

res = norm_axis_angles(a)

# crashes with:
#     res[negative_angle_mask, :3] *= -1.0
# numpy._core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'multiply' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
```

Maybe the type checking is something for the next PR? Probably this is an edge case ...

pytransform3d/batch_rotations.py
```diff
@@ -38,6 +38,52 @@ def norm_vectors(V, out=None):
...
+
+    angle_normalized = norm_angle(angles)
+
+    negative_angle_mask = angle_normalized < 0.0
+    res[negative_angle_mask, :3] *= -1.0
+    angle_normalized[negative_angle_mask] *= -1.0
+
+    res[non_zero_mask, 3] = angle_normalized[non_zero_mask]
+
+    if only_one:
```
kopytjuk, line 82: Consider define the variable next to its usage - during code reading a reader/reviewer has to [search/scroll](https://softwareengineering.stackexchange.com/a/113264) up (ok maybe in this case this is not very critical) ...

```suggestion
    only_one = a.ndim == 1
    if only_one:
```

pytransform3d/batch_rotations.py
```diff
@@ -338,6 +384,45 @@ def axis_angles_from_matrices(Rs, traces=None, out=None):
...
+    -------
+    as : array, shape (..., 4)
+        Axis of rotation and rotation angle: (x, y, z, angle). The angle is
+        constrained to [0, pi) so that the mapping is unique.
+    """
+    quaternion_vector_part = qs[..., 1:]
+    qvec_norm = np.linalg.norm(quaternion_vector_part, axis=-1)
+
+    # Vectorized branches bases on norm of the vector part
+    small_p_norm_mask = qvec_norm < np.finfo(float).eps
```
kopytjuk, line 409: You may consider to include `eps` into `pytransform3d/rotations/_constants.py` and use it from there (at least there is a `eps` instance defined)

pytransform3d/test/test_batch_rotations.py
```diff
@@ -42,6 +43,34 @@ def test_norm_vectors_zero():
     assert_array_almost_equal(V_unit, V)
 
 
+def test_norm_axis_angles():
+    rng = np.random.default_rng(843)
+    A = np.hstack((
+        pbr.norm_vectors(rng.standard_normal(size=(10, 3))),
+        np.pi + rng.uniform(size=(10, 1))
+    ))
```
kopytjuk, line 51: Since you base all your tests on `A`, consider a small refactoring for more clarity (after this the reader knows, "ah yeah those are multiple"):

```suggestion
    # create a batch of axis-angle instances
    num_instances = 10
    A = np.hstack((
        pbr.norm_vectors(rng.standard_normal(size=(num_instances, 3))),
        np.pi + rng.uniform(size=(num_instances, 1))
    ))
```

pytransform3d/test/test_batch_rotations.py
```diff
@@ -42,6 +43,34 @@ def test_norm_vectors_zero():
...
+
+    # 1D
```
kopytjuk, line 53:
```suggestion
    # cross check scalar version with vectorized version
```

pytransform3d/test/test_batch_rotations.py
```diff
@@ -42,6 +43,34 @@ def test_norm_vectors_zero():
...
+    assert_array_almost_equal(
+        pbr.norm_axis_angles(A[0]), pr.norm_axis_angle(A[0]))
+
+    assert_array_almost_equal(
```
kopytjuk, line 57:
```suggestion
    # cross check scalar version with vectorized version when passing an 3D array
    assert_array_almost_equal(
```

pytransform3d/test/test_batch_rotations.py
```diff
@@ -42,6 +43,34 @@ def test_norm_vectors_zero():
...
+        pbr.norm_axis_angles(np.array([[[0.0, 0.0, 0.0, np.pi]]]))[0, 0],
+        pr.norm_axis_angle(np.array([0.0, 0.0, 0.0, np.pi])))
+
+    # 2D
+    A_norm = pbr.norm_axis_angles(A)
```
kopytjuk, line 62: Consider moving the definition of `A` down below, so it it closer to its usage - a reader does not have to scroll up to check what is in A ...

pytransform3d/test/test_batch_rotations.py
```diff
@@ -42,6 +43,34 @@ def test_norm_vectors_zero():
...
+    for a, a_norm in zip(A, A_norm):
+        assert pytest.approx(a[3]) != a_norm[3]
```
kopytjuk, line 64: Consider leaving a comment why is it guaranteed here? A naive reader (like me :D) may assume only the rotation vector is normalized but the angle is unchanged?

pytransform3d/test/test_batch_rotations.py
```diff
@@ -42,6 +43,34 @@ def test_norm_vectors_zero():
...
+        assert_array_almost_equal(a_norm, pr.norm_axis_angle(a))
+
+    # 3D
+    A3D = A.reshape(5, 2, -1)
+    A3D_norm = pbr.norm_axis_angles(A3D)
+    for a, a_norm in zip(A3D.reshape(10, -1), A3D_norm.reshape(10, -1)):
```
kopytjuk, line 70: When you use the `num_instaces` suggestion from above, you get rid of the magic `10` :)

pytransform3d/test/test_batch_rotations.py
```diff
@@ -318,6 +347,28 @@ def test_axis_angles_from_matrices_output_variable():
     assert_array_almost_equal(A, A2_compact)
 
 
+def test_axis_angles_from_quaternions():
+    rng = np.random.default_rng(48322)
+    Q = pbr.norm_vectors(rng.standard_normal(size=(20, 4)))
```
kopytjuk, line 352: Similar case as in `norm_axis_angles` tests:

```suggestion
    num_instances = 20
    Q = pbr.norm_vectors(rng.standard_normal(size=(num_instances, 4)))
```

pytransform3d/test/test_batch_rotations.py
```diff
@@ -318,6 +347,28 @@ def test_axis_angles_from_matrices_output_variable():
...
+
+    # 2D
+    A = pbr.axis_angles_from_quaternions(Q)
+    for a, q in zip(A, Q):
+        pr.assert_quaternion_equal(a, pr.axis_angle_from_quaternion(q))
+
+    # 3D
+    Q3D = Q.reshape(5, 4, 4)
+    A3D = pbr.axis_angles_from_quaternions(Q3D)
+    for a, q in zip(A3D.reshape(20, 4), Q3D.reshape(20, 4)):
```
kopytjuk, line 368:
```suggestion
    for a, q in zip(A3D.reshape(num_instances, 4), Q3D.reshape(num_instances, 4)):
```

pytransform3d/test/test_trajectories.py
```diff
@@ -72,6 +79,39 @@ def test_concat_many_to_one():
     assert_array_almost_equal(A2Cs, concat_many_to_one(A2Bs, B2C))
 
 
+def test_concat_dynamic():
+    rng = np.random.default_rng(84320)
+    A2Bs = np.stack([random_transform(rng) for _ in range(5)])
```
kopytjuk, line 84: For readability:

```suggestion
    num_instances = 5
    A2Bs = np.stack([random_transform(rng) for _ in range(num_instances)])
```

pytransform3d/test/test_trajectories.py
```diff
@@ -72,6 +79,39 @@ def test_concat_many_to_one():
...
+    B2Cs = np.stack([random_transform(rng) for _ in range(5)])
+    A2Cs = concat_dynamic(A2Bs, B2Cs)
+    for i in range(len(A2Cs)):
```
kopytjuk, line 87:
```suggestion
    num_instances = 5
    A2Bs = np.stack([random_transform(rng) for _ in range(num_instances)])
    B2Cs = np.stack([random_transform(rng) for _ in range(num_instances)])
    
    A2Cs = concat_dynamic(A2Bs, B2Cs)
    for i in range(2*num_instances):
```

kopytjuk, line 87: For readability:

```suggestion
    num_instances = 5
    A2Bs = np.stack([random_transform(rng) for _ in range(num_instances)])
    B2Cs = np.stack([random_transform(rng) for _ in range(num_instances)])
    A2Cs = concat_dynamic(A2Bs, B2Cs)
    
    for i in range(num_instances):
```

pytransform3d/test/test_trajectories.py
```diff
@@ -72,6 +79,39 @@ def test_concat_many_to_one():
...
+        assert_array_almost_equal(A2Cs[i], concat(A2Bs[i], B2Cs[i]))
```
kopytjuk, line 88:
```suggestion
        # check the result of the dynamic version equals the one from "one-instance" version
        assert_array_almost_equal(A2Cs[i], concat(A2Bs[i], B2Cs[i]))
```

pytransform3d/test/test_trajectories.py
```diff
@@ -72,6 +79,39 @@ def test_concat_many_to_one():
...
+        assert_array_almost_equal(A2Cs[i], concat_dynamic(A2Bs[i], B2Cs[i]))
+    assert_array_almost_equal(concat_dynamic(A2Bs[0], B2Cs),
+                              concat_one_to_many(A2Bs[0], B2Cs))
+    assert_array_almost_equal(concat_dynamic(A2Bs, B2Cs[0]),
+                              concat_many_to_one(A2Bs, B2Cs[0]))
```
kopytjuk, line 93: Can you give a rough idea of your asserts in a comment? I have a vague idea that you want to check the "dynamic" version of concat against the "one-instance" version ...

pytransform3d/test/test_trajectories.py
```diff
@@ -72,6 +79,39 @@ def test_concat_many_to_one():
...
+    with pytest.raises(ValueError, match="Expected ndim 2 or 3"):
+        concat_dynamic(A2Bs[np.newaxis], B2Cs)
+
+
+def test_concat_many_to_many():
+    rng = np.random.default_rng(84320)
+    A2Bs = np.stack([random_transform(rng) for _ in range(5)])
+    B2Cs = np.stack([random_transform(rng) for _ in range(5)])
+    A2Cs = ptr.concat_many_to_many(A2Bs, B2Cs)
+    for i in range(len(A2Bs)):
```
kopytjuk, line 105:
```suggestion
    num_instances = 5
    A2Bs = np.stack([random_transform(rng) for _ in range(num_instances)])
    B2Cs = np.stack([random_transform(rng) for _ in range(num_instances)])
    A2Cs = ptr.concat_many_to_many(A2Bs, B2Cs)
    for i in range(num_instances):
```

kopytjuk (approved):
Hey guys, you both did an amazing work - the code is nicely written, documented and thoroughly tested!

However I have some suggestions, which we may tackle in future PRs:

- some file sizes are really large, e.g. 800+ lines `pytransform3d/trajectories.py` - we may want to break it down into smaller files - usually smaller files are easier to grasp and navigate
- the unittests of math-heavy function have a lot of asserts, me as a not very math-savvy persion it is hard to assess if they cover everything ... a comment sometimes help a lot what is checked - however maybe it is my own issue

pytransform3d/batch_rotations.py
AlexanderFabisch: No, we want to use the axes, not the angles, but I had to try it to see the mistake, too. :D

pytransform3d/batch_rotations.py
AlexanderFabisch: I would avoid to make the decision in this pull request, since this should be handled consistently in the whole library.

pytransform3d/batch_rotations.py
AlexanderFabisch: We have to do it in the beginning, since the shape of a might change with `a = np.atleast_2d(a)`. The idea is to return only one 1D array, when we pass one 1D array to the function.

pytransform3d/batch_rotations.py
AlexanderFabisch: Yes, but it's a different value. In this case `np.finfo(float).eps`, which is about 1e-16 makes more sense and I'd like to avoid adding multiple eps to the file. I think it makes more sense to remove it from the file altogether.

pytransform3d/test/test_batch_rotations.py
AlexanderFabisch: It's used throughout the test case. I don't know how to solve it better. I don't want to make a new test case for each assert, since I would have to recreate the array in every function.

pytransform3d/test/test_batch_rotations.py
AlexanderFabisch: Good point. Doesn't make sense to normalize the axis when checking this function. I changed that to

```python
    # create a batch of unnormalized axis-angle instances
    n_rotations = 10
    A_unnormalized = np.hstack((
        rng.standard_normal(size=(n_rotations, 3)),
        np.pi + rng.uniform(size=(n_rotations, 1))
    ))
```

Then this loop becomes

```python
    for a_unnormalized, a_norm in zip(A_unnormalized, A_norm):
        # Make sure that the unnormalized version was in fact not normalized.
        assert pytest.approx(a_unnormalized[3]) != a_norm[3]
        assert pytest.approx(np.linalg.norm(a_unnormalized[:3])) != 1.0
        assert_array_almost_equal(a_norm, pr.norm_axis_angle(a_unnormalized))
```

AlexanderFabisch:
> Hey guys, you both did an amazing work - the code is nicely written, documented and thoroughly tested!

Thanks for the review!

>     * some file sizes are really large, e.g. 800+ lines `pytransform3d/trajectories.py` - we may want to break it down into smaller files - usually smaller files are easier to grasp and navigate

Definitely good advice: https://github.com/dfki-ric/pytransform3d/issues/309

With some other large files we have the problem that the dependency structure becomes a bit complicated. For this reason I've avoided this task so far...

>     * the unittests of math-heavy function have a lot of asserts, me as a not very math-savvy persion it is hard to assess if they cover everything ... a comment sometimes help a lot to understand why is something checked - however maybe it is my own issue

So far I did not spend a lot of time explaining the unit tests and they are also not well written to be honest. Would definitely be nice, but not on the top of my priority list. So I could need some help here. Maybe from a not very math-savvy person. :)

AlexanderFabisch:
@JonasHablitzel @kopytjuk One more thing for this PR: does anyone of you maybe have a nice application of the temporal transform manager with a good visualization? I am looking for something to add to the [gallery](https://github.com/dfki-ric/pytransform3d?tab=readme-ov-file#gallery). We don't need to publish the source code for this.

I am also working on a journal paper about pytransform3d at the moment. If you have an interesting example that can be summarized in about half a page, we might also be able to add it there.

AlexanderFabisch:
> Do you envision something in that direction @AlexanderFabisch ?

In principle yes, but I was hoping that you might have interesting data already and it is not too complicated to create a nice visualization. So maybe it's not something for this PR.

AlexanderFabisch:
@kopytjuk @JonasHablitzel Is this PR ready to merge from your perspective?

kopytjuk:
> but I was hoping that you might have interesting data already

I think going through the interal processes of my employer to expose internal data to the public is definitely not in time horizon of this PR :D

> Is this PR ready to merge from your perspective?

yes!

AlexanderFabisch:
> I think going through the interal processes of my employer to expose some interesting internal data to the public is definitely not in time horizon of this PR 😄

If it's easier, I'm just asking for a nice picture, not for the data, but I guess it's the same issue.

JonasHablitzel:
yeah agree on the idea with the robot. I don't have somthing like this currently but maybe we could use a open dataset like [kitti-360](https://www.cvlibs.net/datasets/kitti-360/) or [kitti-360](https://www.cvlibs.net/datasets/kitti-360/) or [pandaset](https://www.kaggle.com/datasets/usharengaraju/pandaset-dataset) i can prepare a small example/picture

AlexanderFabisch:
OK, for now I would merge the PR and make a new release of pytransform3d. Thanks @JonasHablitzel for this great contribution and thanks @kopytjuk for the review.

JonasHablitzel:
i also have to thank you guys for the great support :)
here is already a preview of the kitti360 dataset complety ego motion compesated + transformed into a global frame without any c-code [kitti360Scripts](https://github.com/autonomousvision/kitti360Scripts/tree/master/kitti360scripts/devkits/accumuLaser). I wanted also compare the pytransform3d implementation against this c++ implementation regarding speed. I think i will finish it by the end of the week

image

AlexanderFabisch:
> I wanted also compare the pytransform3d implementation against this c++ implementation regarding speed. I think i will finish it by the end of the week

Vectorization has to be really good to achieve a comparable speed. But maybe we can improve it a little bit when you did the benchmark.

Another option that I have on my agenda for some time already is to think about making the code compatible with the bumpy interface of Jax, which would give us jit compilation and GPU support for parallelization. Maybe we could start by making the operations required in this use case compatible.

AlexanderFabisch:
I just released pytransform3d, which includes this feature. :)

Issue #350, opened 2025-06-21
Add support for scaling size attributes to set_data.


ShimonWang:
When I created a `Figure` object and drew a mesh using the `plot_mesh` method, I wanted to update the zoom level of the mesh, but this property interface does not seem to be available in `set_data`.
![Image](https://github.com/user-attachments/assets/bf387037-d275-42fb-9b12-d026a810698b)

![Image](https://github.com/user-attachments/assets/b7cc62f4-08f3-4e25-960a-5e727e35a6ef)

![Image](https://github.com/user-attachments/assets/61cec5fe-7178-4b2c-8eff-a534af11925b)

AlexanderFabisch:
Hi @ShimonWang ,

so far I did not see the need to add it. It is a bit unusual to resize the mesh in an animation. Is `Figure.set_zoom(...)` not sufficient in your case?

It should not be too hard to integrate it in a way similar to how it is done in the constructor. Would you like to make a pull request for this?

https://github.com/dfki-ric/pytransform3d/blob/91091922bd68abc52a412962b8ba23ff1d03f65d/pytransform3d/visualizer/_artists.py#L593-L627

ShimonWang:
Thank you for your suggestion. I initially tried copying the original mesh object, but this resulted in unnecessary memory increase. I tried to correctly convert the scale size based on your idea of updating the A2B pose. I will pull a request for this issue. Please review it.

PR #352, opened 2025-06-22
Add support for scaling factor s in set_data

ShimonWang:
I tried to add a feature to update the size of mesh support. https://github.com/dfki-ric/pytransform3d/issues/350

### DIFF
--- a/pytransform3d/visualizer/_artists.py
+++ b/pytransform3d/visualizer/_artists.py
@@ -597,9 +597,10 @@ class Mesh(Artist):
         if convex_hull:
             mesh.convex_hull()
 
+        self.s = s
         self.mesh = mesh.get_open3d_mesh()
         self.mesh.vertices = o3d.utility.Vector3dVector(
-            np.asarray(self.mesh.vertices) * s
+            np.asarray(self.mesh.vertices) * self.s
         )
         self.mesh.compute_vertex_normals()
         if c is not None:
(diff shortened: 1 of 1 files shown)


pytransform3d/visualizer/_artists.py
```diff
@@ -610,14 +611,25 @@ def __init__(
...
         A2B : array-like, shape (4, 4)
             Center of the mesh.
+
+        s : array-like, shape (3,), optional (default: [1, 1, 1])
+            Scaling of the mesh that will be drawn
         """
+        previous_s = self.s
+        if previous_s is None:
+            previous_s = np.ones(3)
+        self.mesh.vertices = o3d.utility.Vector3dVector(
```
AlexanderFabisch, line 628: The problem with this approach is that the result depends on the mesh being in the original coordinate frame. The scaling will be wrong if it is not. That's why I suggested to keep a copy of the original mesh.

You could work around that problem by first "undoing" the transformation `previous_A2B`, then scaling the mesh, and finally transforming it with A2B, but this will most likely cause issues with numerical precision at some point.

I'd suggest to make this new behavior optional. It should be possible to turn it on or off in the constructor. Then you can keep a copy of the original mesh and scale it as needed in `set_data`. Would that work for you?

ShimonWang: Thank you for your reply and suggestions, but this solution seems to complicate the original method. Furthermore, with so many `Artist` objects in the project using the `set_data` method, the benefits seem to be far outweighed by the costs, and it may also lead to project instability. I will consider other methods to achieve a similar effect in the examples I need, rather than directly modifying the `pytransform3d` code itself.

PR #355, opened 2025-09-05
Fix for URDF parsing with XML namespace

fagg:
The current parser does not properly handle URDF files which contain an `xmlns` (XML namespace) attribute. This is largely to do with the underlying XML parser which successfully parses the tree but prepends the tag/name for all elements (incl root) with `{namespace}`. This breaks the assumptions made about the root tag and causes the parser to raise an exception.

Additionally, further nodes will not parse correctly with this namespace tag.

To resolve this:

1. Add a normalization step which traverses the graph and strips the namespace prefix from all elements
2. Adds a test with an example XML string.

It was surprising to me that the parser fails when it is particularly common to see URDF files with a namespace (most commonly from ROS). I am not sure of a better/cleaner way, but a simple normalization (i.e. removal of the namespace prefix) seems logical.

Tested with Python 3.12 on Windows. The test suite passes successfully.

### DIFF
--- a/pytransform3d/test/test_urdf.py
+++ b/pytransform3d/test/test_urdf.py
@@ -191,6 +191,36 @@ def test_missing_child_link_name():
         UrdfTransformManager().load_urdf(urdf)
 
 
+def test_xml_namespace():
+    urdf = """
+    <robot name="robot_name" xmlns="http://www.ros.org">
+    <link name="link0"/>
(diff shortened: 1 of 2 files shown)


AlexanderFabisch:
Hi @fagg, thanks for the contribution. I merged it to the develop branch and it will be part of the next release on PyPI.

PR #362, opened 2026-05-07
[MRG] Viser visualization backend

AlexanderFabisch:
### `pytransform3d.viser` — web-based visualization backend

A new `pytransform3d.viser` module backed by [viser](https://viser.studio/) that is a drop-in replacement for `pytransform3d.visualizer`. It renders in a browser via WebSocket, so it works in headless environments (remote servers, containers, notebooks) where an OpenGL window is unavailable. Switching backends is a one-line import change; both modules share the same public API.

Differences from the Open3D backend: `show()` is non-blocking and prints a URL instead of opening a window; `save_image()` raises `NotImplementedError`; `set_line_width()` is a no-op. `view_init()` gains `center` and `distance` parameters for controlling the orbit target.

Install with `pip install 'pytransform3d[all]'` or `pip install viser`. `trimesh` is required for shapes without a native viser primitive (cylinders, cones, capsules, ellipsoids, vector arrows) and is included in the `all` extras group.

### New examples

Six new examples in `examples/viser/`:

- `vis_viser_shapes` — all geometric primitives in one static scene
- `vis_viser_robot_arm` — animated 6-DOF robot arm with growing TCP trajectory
- `vis_viser_wrench_dynamics` — rigid-body simulation driven by a body-fixed wrench with rolling position trail
- `vis_viser_camera_orbit` — pinhole camera orbiting a static scene, frustum updated every frame
- `vis_viser_uncertain_transforms` — banana distribution from SE(3) uncertainty propagation: MC paths, mean trajectory, projected hyperellipsoid, and Cartesian position ellipsoid
- `vis_viser_probabilistic_robot_kinematics` — PPOE probabilistic forward kinematics with animated end-effector pose-uncertainty ellipsoid

A new user-guide page covers installation, the API, differences from the Open3D backend, and the animation workflow.

### DIFF
--- a/pytransform3d/viser/_artists.py
+++ b/pytransform3d/viser/_artists.py
@@ -1243,13 +1243,74 @@ class Camera(Artist):
         self.strict_check = strict_check
         self._handle = None
 
-    def _fov_and_aspect(self):
-        fy = self.M[1, 1]
-        h = self.sensor_size[1]
-        w = self.sensor_size[0]
-        fov = 2.0 * np.arctan(h / (2.0 * fy))
-        aspect = float(w) / float(h)
-        return fov, aspect
+    def _compute_segments(self):
+        """Compute the 11 line segments that form the camera wireframe.
+
+        The wireframe consists of four frustum rays from the camera centre to
+        the corners of the virtual image plane, the rectangle connecting those
+        corners, and a small triangle above the top edge that indicates the
+        camera's up direction.
+
+        Returns
+        -------
+        segments : array, shape (11, 2, 3)
+            Line segments in world coordinates.
+        """
+        cam2world = self.cam2world
+        focal_length = float(np.mean([self.M[0, 0], self.M[1, 1]]))
+        w, h = float(self.sensor_size[0]), float(self.sensor_size[1])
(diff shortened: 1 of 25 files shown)


pytransform3d/viser/_artists.py
```diff
@@ -0,0 +1,1579 @@
...
+    color : tuple of int or None
+        RGB color with values in [0, 255], or None.
+    """
+    if c is None:
+        return None
+    arr = np.asarray(c, dtype=float)
+    return tuple(int(np.clip(v * 255.0, 0, 255)) for v in arr[:3])
+
+
+def _wxyz_from_matrix(A2B):
```
AlexanderFabisch, line 36: Should be inlined

pytransform3d/viser/_artists.py
```diff
@@ -0,0 +1,1579 @@
...
+
+    Returns
+    -------
+    wxyz : array, shape (4,)
+        Quaternion in wxyz convention.
+    """
+    return pr.quaternion_from_matrix(A2B[:3, :3], strict_check=False)
+
+
+def _make_arrow_mesh(length):
```
AlexanderFabisch, line 52: Not tested yet

PR #363, opened 2026-05-19
Numerical code audit

AlexanderFabisch:


pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -18,7 +18,7 @@ def norm_axis_angles(a):
     -------
     a : array, shape (..., 4)
         Axis of rotation and rotation angle: (x, y, z, angle). The length
-        of the axis vector is 1 and the angle is in [0, pi). No rotation
+        of the axis vector is 1 and the angle is in [0, pi]. No rotation
```
clementperon, line 21: Was reading the doc and about to open a MR for this issue 👍

clementperon: As there could be two representation for this maybe it would be nice to force a deterministic axis sign no?

AlexanderFabisch: Hi, yes, this makes sense. Thanks for the suggestion. What would you suggest as a solution? Would you like to open a pull request to the develop branch? We also have to make sure that rotations.norm_axis_angle is consistent.

clementperon: The rule I have set in my repo is
"
At `angle = 180°`, `axis` and `-axis` represent the same rotation. To make the representation unique, the first non-zero component of `axis` SHALL be positive.
"

AlexanderFabisch: Would you like to open a pull request for this? If not, I will do it when I have time.

clementperon: I would be glad to contribute but I have finally implemented it natively without using pytransform3d lib (that's the only feature I was looking for)
As I have zero knowledge in this repo, sure an AI agent will do better than me 😄

AlexanderFabisch:
Merged with 1a9eb5bc465e8e8d7bac3ecb01931204f88483a8

PR #367, opened 2026-07-09
[Rotation] Swing & Twist Decomposition

fagg:
Adds two new functions to pytransform3d.rotations:

- `swing_twist_decomposition` - splits a unit quaternion into a twist (rotation about a given axis) and a swing (rotation about an axis orthogonal to it), such that `q = q_swing . q_twist`.
- `swing_twist_composition` - the inverse operation, for reconstruction of the original rotation from its swing and twist components.

This decomposition is useful for enforcing joint limits, separating roll about a link axis from the remaining rotation about a symmetry axis. It follows Dobrowolski (2015), "Swing-twist decomposition in Clifford algebra" - https://arxiv.org/abs/1506.05481

Algorithmic details:

- Twist is obtained by projecting the vector part of the quaternion onto the twist axis and renormalizing.
- The swing is recovered as `q_swing = q . q_twist^-1`
- Quaternions doubly cover SO(3) and this surfaces in two places. The first is that decomposition is sign invariant - as such the twist is canonicalized into the hemisphere with non-negative scalar part (w >= 0), so its rotation angle about the axis stays in [-pi, pi]. As a result, `swing_twist_decomposition(q, axis)` and `swing_twist_decomposition(-q, axis)` return the exact same twist quaternions and equivalent swings. Secondly, composition may cause a sign flip - you may get `-q` instead of `q`.  

Implementation details:

- A zero twist axis will raise a ValueError
- The twist axis is rescaled by its largest-magnitude component before normalization, so subnormal axes don't underflow to zero (leading to a spurious zero-error vector) or lose precision.
- For a rotation of pi about an axis orthogonal to the twist axis, the twist is undefined (scalar part and axis projection both vanish). In this case we set the twist to identity and swing carries the full rotation.

All cases and constraints described above are covered by unit tests.

### DIFF
--- a/doc/source/user_guide/rotations.rst
+++ b/doc/source/user_guide/rotations.rst
@@ -398,6 +398,59 @@ with the quaternion product (:func:`~pytransform3d.rotations.q_prod_vector`).
 * Interpretation: not straightforward.
 * Ambiguities: double cover.
 
+Swing-Twist Decomposition
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Any rotation can be split into a *twist* about a chosen axis
+:math:`\boldsymbol{e}` and a *swing* that carries the remaining orientation,
+such that
+
+.. math::
+
+    \boldsymbol{q} = \boldsymbol{q}_{swing} \boldsymbol{q}_{twist},
+
+where the twist is applied first. The twist
+:math:`\boldsymbol{q}_{twist}` is a rotation about :math:`\boldsymbol{e}`
+and the swing :math:`\boldsymbol{q}_{swing}` is a rotation about an axis that
+is orthogonal to :math:`\boldsymbol{e}` [9]_. The twist is obtained by
+projecting the vector part of the quaternion onto the twist axis and
+renormalizing; the swing is then recovered as
+:math:`\boldsymbol{q}_{swing} = \boldsymbol{q} \boldsymbol{q}_{twist}^{-1}`.
+
+This is useful to separate the roll about a link axis from the rest of an
+orientation, to enforce joint limits, or to isolate the rotation about a
+symmetry axis. pytransform3d provides
+:func:`~pytransform3d.rotations.swing_twist_decomposition` and the inverse
+:func:`~pytransform3d.rotations.swing_twist_composition`, which reconstructs
+the original rotation from swing and twist.
+
+.. plot::
+    :include-source:
+
+    import numpy as np
+    from pytransform3d.rotations import (
+        swing_twist_decomposition, quaternion_from_axis_angle,
+        matrix_from_quaternion, plot_basis)
+
+    axis = np.array([0.0, 0.0, 1.0])
+    q = quaternion_from_axis_angle([0.3, 0.7, 0.2, 1.2])
+    swing, twist = swing_twist_decomposition(q, axis)
+
+    ax = plot_basis(R=matrix_from_quaternion(q), p=[0, 0, 0])
+    plot_basis(ax=ax, R=matrix_from_quaternion(twist), p=[3, 0, 0])
+    plot_basis(ax=ax, R=matrix_from_quaternion(swing), p=[6, 0, 0])
+
+The left frame shows the full rotation, the middle frame the twist (a pure
+rotation about the z-axis), and the right frame the swing.
+
+.. warning::
+
+    The twist is undefined for a rotation by :math:`\pi` about an axis
+    orthogonal to the twist axis: both the scalar part and the projection onto
+    the twist axis vanish. In this degenerate case
+    :func:`~pytransform3d.rotations.swing_twist_decomposition` returns the
+    identity as the twist and lets the swing carry the full rotation.
+
 ------------
 Euler Angles
 ------------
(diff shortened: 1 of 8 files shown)


AlexanderFabisch:
Hi @fagg ,

thank you for this contribution too!

A few comments:

1. Please run black + ruff check
2. Should describe it in the user guide? Maybe to Rotations -> Quaternions?
3. You say

> This decomposition is useful, for example, to enforce joint limits, to separate the roll about a link axis from the remaining orientation, or to isolate the rotation about a symmetry axis.

It would be really interesting to have at least one example that demonstrates one of these use cases. I'd prefer the joint limit use case, since this is most interesting to me, but it does not have to be this one. Would it be possible to add one example?

4. The implementation of the decomposition adds conventions beyond the algorithm from the paper that should be clearly documented: (1) `twist_norm < eps` and (2) `twist[0] < 0`.

fagg:
Hello @AlexanderFabisch,

>     1. Please run black + ruff check

Done.

>     2. Should describe it in the user guide? Maybe to Rotations -> Quaternions?

I added a draft set of changes.

>     3. 
> It would be really interesting to have at least one example that demonstrates one of these use cases. I'd prefer the joint limit use case, since this is most interesting to me, but it does not have to be this one. Would it be possible to add one example?

I have added two.

One is simple plot - which shows the original rotation, it's decomposition into twist and swing, and it's composition from twist and swing.

The more complex of the two attempts to demonstrate axis redundancy using this decomposition.

What it shows is a robot end-effector holding a tool whose working axis is the body z-axis, where the tool is pointing towards a workpiece. It builds a nominal porientation (aim + arbitrary roll), decomposes it about the tool axis into swing (where the tool axis points) and twist (roll about the tool). We then visualize the sweep along the twist while holding the swing fixed. Effectively, what we generated is a set of end-effector orientations that keep the tool pointed at the workpiece - you could consider this a solution set for a task-space planner with twist being the only remaining free parameter.

>     4. The implementation of the decomposition adds conventions beyond the algorithm from the paper that should be clearly documented: (1) `twist_norm < eps` and (2) `twist[0] < 0`.

I already had tried to make that clear in the comments - it seems I failed. The latest version here has beefed up language around those assumptions in the doc string.

AlexanderFabisch:
This looks really good. Thanks! I will test everything next week.

examples/plots/plot_swing_twist.py
```diff
@@ -0,0 +1,78 @@
+"""
+==========================
```
AlexanderFabisch, line 2: Minor detail, but these `===`s should have the same length as the title.

examples/plots/plot_swing_twist_tool_axis.py
```diff
@@ -0,0 +1,116 @@
+"""
+=========================================
```
AlexanderFabisch, line 2: Same here

doc/source/user_guide/rotations.rst
```diff
@@ -398,6 +398,59 @@ with the quaternion product (:func:`~pytransform3d.rotations.q_prod_vector`).
 * Interpretation: not straightforward.
 * Ambiguities: double cover.
 
+Swing-Twist Decomposition
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
AlexanderFabisch, line 402: `~`s should have same length as title

doc/source/user_guide/rotations.rst
```diff
@@ -398,6 +398,59 @@ with the quaternion product (:func:`~pytransform3d.rotations.q_prod_vector`).
...
+
+Any rotation can be split into a *twist* about a chosen axis
+:math:`\boldsymbol{e}` and a *swing* that carries the remaining orientation,
+such that
+
+.. math::
+
+    \boldsymbol{q} = \boldsymbol{q}_{swing} \boldsymbol{q}_{twist},
```
AlexanderFabisch, line 410: Could you use `\text{swing}` and `\text{twist}` here and in the docstrings?

doc/source/user_guide/rotations.rst
```diff
@@ -398,6 +398,59 @@ with the quaternion product (:func:`~pytransform3d.rotations.q_prod_vector`).
...
+symmetry axis. pytransform3d provides
+:func:`~pytransform3d.rotations.swing_twist_decomposition` and the inverse
+:func:`~pytransform3d.rotations.swing_twist_composition`, which reconstructs
+the original rotation from swing and twist.
+
+.. plot::
+    :include-source:
+
+    import numpy as np
+    from pytransform3d.rotations import (
```
AlexanderFabisch, line 431: Could you also use black-like formatting here?

AlexanderFabisch:
Had time to review already now. It's very well documented. I just have some comments on little details. This is a really good contribution, thanks!

AlexanderFabisch:
I applied the cosmetic changes and merged this feature to develop. It will be part of release 3.16.0 of pytransform3d. Thanks a lot for the contribution!

fagg:
Thanks, sorry for the delay in getting back to those suggestions. I will get back to you on https://github.com/dfki-ric/pytransform3d/pull/368 sometime this week hopefully.

AlexanderFabisch:
No worries. Thanks.

PR #368, opened 2026-07-09
[Rotation] Add 6D rotation representation per Zhou et al 2019

fagg:
Adds two new functions to pytransform3d.rotations for converting between rotation matrices and continuous 6D rotation representation introduced by Zhou et al (CVPR 2019).

- `rotation_6d_from_matrix` - encode a rotation matrix as a 6D vector (its first two columns stacked)
- `matrix_from_rotation_6d` - decode a 6D vector back into a rotation matrix via Gram-Schmidt orthonormalization

Most parameterizations of rotation (quaternion, axis-angle, Euler angles) are discontinuous as functions on SO(3). For classical geometry, this largely doesn't matter. However, the discontiuity is problematic for machine learning. Zhou et al showed that any representation of SO(3) with fewer than 5 dimensions is necessarily discontinuous - and proposed a continuous 6D representation that avoids this. It has become the defacto standard target for learned rotation estimation in pose estimation and other computer vision tasks. This PR makes this representation directly available in pytransform3d.

All of the core functionality already existed, this merely wraps it. A proper test suite has been added to ensure correctness;


AlexanderFabisch:
Hi @fagg ,

thank you for this contribution. I only have two open points:

1. Representation name: not sure how much you thought about it yet. Do you think `rotation_6d` is the best name for this? Which name is used in the paper? Is it intuitive? Is there maybe a better name? We already have `matrix_from_two_vectors`, where `two_vectors` is the same representation, but arranged in a different way. We should clearly distinguish both. I don't have any better idea at the moment, but I'd like to at least have a brief discussion about this.

2. Extension of the user guide: it is not strictly necessary to add the representation everywhere, but it could be added to

* overview figure: https://dfki-ric.github.io/pytransform3d/user_guide/introduction.html#representations
* rotations section: https://dfki-ric.github.io/pytransform3d/user_guide/rotations.html
    * figure
    * table
    * its own section
* we could also add another section about good representations for machine learning

What do you think?

fagg:
> Hi @fagg ,
> 
> thank you for this contribution. I only have two open points:
> 
>     1. Representation name: not sure how much you thought about it yet. Do you think `rotation_6d` is the best name for this? Which name is used in the paper? Is it intuitive? Is there maybe a better name? We already have `matrix_from_two_vectors`, where `two_vectors` is the same representation, but arranged in a different way. We should clearly distinguish both. I don't have any better idea at the moment, but I'd like to at least have a brief discussion about this.
> 
>     2. Extension of the user guide: it is not strictly necessary to add the representation everywhere, but it could be added to
> 
> 
>     * overview figure: https://dfki-ric.github.io/pytransform3d/user_guide/introduction.html#representations
> 
>     * rotations section: https://dfki-ric.github.io/pytransform3d/user_guide/rotations.html
>       
>       * figure
>       * table
>       * its own section
> 
>     * we could also add another section about good representations for machine learning
> 
> 
> What do you think?

So regarding naming - I did a bit of a deep dive on this. THe paper only ever calls it the "6D representation". Simiarly, Pytorch3D's public API is `rotation_6d_to_matrix` and `matrix_to_rotation_6d`.It's distinct from `two_vectors` in intent - `matrix_from_two_vectors` is a geometric constructor from two meaningful directions, whereas the 6D form is an opaque (but learnable) encoding.

For the user guide: I've added the represetnation to both overview figures plus added a table row and a dedicated 6D represetnation section with references to the paper.

Hope that helps, let me know what you think.

doc/source/_static/overview.png
AlexanderFabisch, line 1: The arrow label is overlapping with another label. Could you fix this?

doc/source/_static/rotations.png
AlexanderFabisch, line 1: The  formatting of the arrow head is strange here.

AlexanderFabisch (changes requested):

AlexanderFabisch:
Thank you. Just two minor comments.

doc/source/_static/overview.png
fagg: Fixed

doc/source/_static/rotations.png
fagg: Fixed

AlexanderFabisch:
Thank you for this contribution! It will be part of the next release.

PR #370, opened 2026-07-20
Fix axis-angle extraction for a general axis at pi

gaoflow:
At θ=π the rotation matrix is symmetric, so `axis_angle_from_matrix` currently derives the axis-component signs from `np.sign()` of the skew-symmetric part `R - Rᵀ`, which is numerically zero — just floating-point noise. For a general (non-coordinate) axis this returns a wrong axis (a *different* rotation, not a valid ±negation): about 40% of random π rotations come out with one or more component signs flipped, with round-trip error up to ~1.5.

#364 fixed only the coordinate-axis sub-case in the batch `axis_angles_from_matrices` (where the off-diagonal components are exactly zero), and its `test_axis_angles_from_matrices_near_pi` covers only coordinate axes. The single-matrix `axis_angle_from_matrix` was never touched by it and is still affected — as is the batch function for a general axis.

Since `R = 2 eeᵀ − I` at π, `eeᵀ = ½(R + I)` and the signs live in the symmetric part. This recovers them from the dominant row of `eeᵀ`, using the symmetric part of `R` so the result stays accurate just below π, where the skew part then fixes the overall axis sign. Both functions are fixed for arbitrary axes; coordinate-axis π (including #364's cases) and the near-π path are unchanged.

Everything built on the log map inherits the fix — `compact_axis_angle_from_matrix`, SE(3) log, `matrix_slerp` through 180°, trajectory exponential coordinates and the SO(3)/SE(3) Fréchet mean.

Verified by round-trip and against `scipy.spatial.transform.Rotation`. Added general-axis π regression tests (random axes, single and batch) that round-trip the recovered axis-angle back to the input matrix, completing #364's coordinate-only coverage.

### DIFF
--- a/pytransform3d/batch_rotations/_matrix.py
+++ b/pytransform3d/batch_rotations/_matrix.py
@@ -50,32 +50,35 @@ def axis_angles_from_matrices(Rs, traces=None, out=None):
 
     # The threshold is a result from this discussion:
     # https://github.com/dfki-ric/pytransform3d/issues/43
-    # The standard formula becomes numerically unstable, however,
-    # Rodrigues' formula reduces to R = I + 2 (ee^T - I), with the
-    # rotation axis e, that is, ee^T = 0.5 * (R + I) and we can find the
-    # squared values of the rotation axis on the diagonal of this matrix.
-    # We can still use the original formula to reconstruct the signs of
-    # the rotation axis correctly.
+    # The standard formula becomes numerically unstable near pi.
     angle_close_to_pi = np.abs(angles - np.pi) < 1e-4
     angle_zero = angles == 0.0
     angle_not_zero = np.logical_not(angle_zero)
 
-    Rs_diag = np.einsum("nii->ni", Rs.reshape(-1, 3, 3))
-    if instances_shape:
-        Rs_diag = Rs_diag.reshape(*(instances_shape + (3,)))
-    else:
-        Rs_diag = Rs_diag[0]
+    if np.any(angle_close_to_pi):
+        # ee^T = 0.5 * (R + I) has the squared axis components on its
+        # diagonal. At exactly pi the skew part R - R^T (out[..., :3]) is
+        # numerically zero, so its sign cannot recover the axis for a general
+        # (non-coordinate) axis. We take the relative signs from the dominant
+        # row of ee^T instead, using the symmetric part of R to stay accurate
+        # just below pi, where the skew part then fixes the overall axis sign.
+        Rs_pi = Rs[angle_close_to_pi]
+        Rs_pi_sym = 0.5 * (Rs_pi + np.swapaxes(Rs_pi, -1, -2))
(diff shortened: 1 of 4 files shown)


AlexanderFabisch:
Hi, thank you for this contribution. Could you briefly describe in math what the code is doing so that I can verify it more easily?

binggao1230:
Sure. Here is the math, keyed to the changed lines in `axis_angle_from_matrix` (`rotations/_matrix.py`); the batch path in `batch_rotations/_matrix.py` does the identical thing element-wise per instance.

**Rodrigues at θ = π.** With unit axis e,

    R = I + sin θ [e]× + (1 − cos θ) [e]×²,   [e]×² = eeᵀ − I.

At θ = π (sin π = 0, 1 − cos π = 2) this collapses to

    R = 2 eeᵀ − I   ⟹   eeᵀ = ½(R + I).

**Magnitudes (the diagonal).** eᵢ² = ½(R_ii + 1), so |eᵢ| = √(½(R_ii + 1)). That is `eeT_diag = clip(0.5·(diag(R_sym) + 1), 0, 1)`; the `clip` only guards floating-point drift outside [0, 1].

**Relative signs (the off-diagonal).** (eeᵀ)_ij = eᵢeⱼ. Pick the dominant component k = argmax |eᵢ| (`k = np.argmax(eeT_diag)`). Row k of the symmetric part gives, up to the positive factor 2, sign((R_sym)_kj) = sign(eₖeⱼ) = sign(eₖ)·sign(eⱼ). So with `signs = np.sign(R_sym[k])` and `signs[k] = 1`,

    a[:3] = √(eeT_diag) · signs = sign(eₖ) · e,

i.e. e up to a single overall sign, anchored so the dominant component is positive. Anchoring on the largest |eᵢ| avoids resolving the signs through a near-zero component. I use the symmetric part `R_sym = ½(R + Rᵀ)` rather than R itself so that, just below π, those off-diagonal entries isolate the eᵢeⱼ term (the diagonal is unchanged since (R + Rᵀ)_ii = 2 R_ii).

**Overall sign.** At exactly θ = π, e and −e give the same R, so the overall sign is a genuine gauge freedom. For θ strictly below π it is not, and the skew part

    axis_unnormalized = [R21−R12, R02−R20, R10−R01] = 2 sin θ · e

still carries the correct direction, so

    if angle < π and dot(a[:3], axis_unnormalized) < 0: a[:3] = -a[:3]

flips the recovered axis to agree with it.

**What was wrong before.** The old code took the signs directly from `sign(axis_unnormalized)`, i.e. from R − Rᵀ, which is numerically zero at π — for a general axis that zeros/randomizes the signs. #364 only covered coordinate axes, whose off-diagonal eᵢeⱼ vanish, so the defect never surfaced there.

AlexanderFabisch:
Thanks for the contribution! I merged it to develop and it will be released with version 3.16.0.

Issue #43, opened 2020-04-24
rotations.axis_angle_from_matrix() inaccuracy at angles close to zero or pi


mojodo:
The `y=arccos(x)` function results in numerical imprecision when y should be close to 0 or pi, i.e. when x is close to 1 or x is close to -1.

To circumvent this you could for example write the following:

    # determine angle
    angle = np.arccos((np.trace(R) - 1.0) / 2.0)
        
    if angle < 0.001: # get numerical precision at low angular values
       angle = np.arcsin(np.linalg.norm(R-np.eye(3))/np.sqrt(2))
       if angle == 0.0:
            return np.array([1.0, 0.0, 0.0, 0.0])

    a = np.empty(4)
    D = R-R.transpose()
  
    if angle > np.pi-0.001:# get numerical precision at high angular values
        angle = np.pi - np.arcsin(np.linalg.norm(D)/np.sqrt(8))
    
    # determin rotation axis
    r = np.array([D[2, 1], D[0, 2], D[1, 0]])
    # The norm of r is 2.0 * np.sin(angle)
    a[:3] = r / (2.0 * np.sin(angle))
    a[3] = angle
    return a

AlexanderFabisch:
Hi @mojodo ,

thanks for your feedback.

> The `arccos(x)` function results in numerical imprecision when x is close to 0 or pi.

I am confused. arccos is only defined for -1 <= x <= 1.

Do you have an example of which operation fails and how?

mojodo:
Thanks for asking back. You can try out with any rotation Matrix whoes rotation angle is close to 0 or close to pi. 
Another issue in your code is the special case that you create when angle is close to pi. In this case you compute the components of the rotation axis by some positive square roots. By this you implicitly assume that the signs of the three components are all either positive or negative. However  in general their signs may be different. That is why I suggested the method above.

To ensure that a function works properly in all situations it is good advice, to test them with the most critical cases. for this function this would be, rotation by angle=0 and 0+ epsilon, rotation by angle pi and pi-epsilon, rotation arround axis with differently signed components. This can be done by a set of say 1000 realizations of random rotation matrices spanning the entire space of possible rotations.

To be true, I did not make a rigorous test for my proposed function above.

AlexanderFabisch:
> Thanks for asking back. You can try out with any rotation Matrix whoes rotation angle is close to 0 or close to pi.

[...]

PR #372, opened 2026-08-14
Fix non-deterministic axis for 180-deg rotations (#366)

### LINKED ISSUE #366: norm_axis_angles and norm_axis_angle should produce deterministic axis for 180 degree rotations
See discussion in #363 

> At angle = 180°, axis and -axis represent the same rotation. To make the representation unique, the first non-zero component of axis SHALL be positive.

TabishShahMohsin:
Ensure norm_axis_angle and norm_axis_angles always return a deterministic rotation axis for angles of exactly pi by flipping negative leading non-zero vector components. Adds unit tests.

Fixes #366


AlexanderFabisch:
Hi @TabishShahMohsin , thanks for the contribution. I don't know why copilot decided to do the review, but I will start mine now. :)

pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -45,6 +45,19 @@ def norm_axis_angles(a):
 
     res[rot_mask, 3] = angle_normalized[rot_mask]
 
+    # Issue #366: Make axis deterministic for 180 degree rotations
+    # the first non-zero component of axis should be positive.
+    pi_mask = np.isclose(res[..., 3], np.pi) & rot_mask
```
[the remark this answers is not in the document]
AlexanderFabisch: This is what I thought too. Please change the condition. `isclose` matches too broadly.

pytransform3d/rotations/_axis_angle.py
```diff
@@ -94,6 +94,18 @@ def norm_axis_angle(a):
 
     res[3] = angle
 
+    # Issue #366: Make axis deterministic for 180 degree rotations
+    # the first non-zero component of axis should be positive.
+    if np.isclose(angle, np.pi):
```
AlexanderFabisch: Same here. `isclose` matches to broadly. The modification should just apply to exactly pi.

pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -45,6 +45,19 @@ def norm_axis_angles(a):
...
+    if np.any(pi_mask):
+        axes = res[pi_mask, :3]
+        is_zero = np.isclose(axes, 0.0)
+        first_non_zero_idx = np.argmax(~is_zero, axis=1)
```
AlexanderFabisch: 1. Can we rename this to `first_non_zero_idx_per_entry` or something similar?
2. What if the input is not a 2d array, but generally an N-dimensional array, e.g., an array of an array of axis-angle representations? `axis=1` should be replaced by `axis=-1`.
3. What if the axis is `[0, 0, 0]`? This is an edge case.

pytransform3d/batch_rotations/test/test_batch_rotations.py
```diff
@@ -659,3 +659,18 @@ def test_smooth_quaternion_trajectory_empty():
         ValueError, match=r"At least one quaternion is expected"
     ):
         pbr.smooth_quaternion_trajectory(np.zeros((0, 4)))
+
+
+def test_norm_axis_angle_180_degrees_deterministic_batch():
+    A = np.array([
```
AlexanderFabisch, line 671: I'd add additional tests for [0, 0, 0] and random axes.

pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -45,6 +45,19 @@ def norm_axis_angles(a):
...
+        row_indices = np.arange(len(axes))
```
AlexanderFabisch: Why don't we just use `:` instead of `np.arange`?

pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -45,6 +45,19 @@ def norm_axis_angles(a):
...
+        first_non_zero_vals = axes[row_indices, first_non_zero_idx]
+        flip_mask = first_non_zero_vals < 0.0
```
AlexanderFabisch: I think we don't need an additional variable. I think `flip_mask` is harder to understand than `first_non_zero_vals < 0.0`.

pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -45,6 +45,19 @@ def norm_axis_angles(a):
...
 
+    # Issue #366: Make axis deterministic for 180 degree rotations
+    # the first non-zero component of axis should be positive.
+    pi_mask = np.isclose(res[..., 3], np.pi) & rot_mask
+    if np.any(pi_mask):
+        axes = res[pi_mask, :3]
+        is_zero = np.isclose(axes, 0.0)
+        first_non_zero_idx = np.argmax(~is_zero, axis=1)
+        row_indices = np.arange(len(axes))
+        first_non_zero_vals = axes[row_indices, first_non_zero_idx]
```
AlexanderFabisch: `non_zero_vals` is a bit generic. Can we find a better name for this variable? Maybe `first_nonzero_axis_components`?

pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -45,6 +45,19 @@ def norm_axis_angles(a):
 
     res[rot_mask, 3] = angle_normalized[rot_mask]
 
+    # Issue #366: Make axis deterministic for 180 degree rotations
+    # the first non-zero component of axis should be positive.
+    pi_mask = np.isclose(res[..., 3], np.pi) & rot_mask
+    if np.any(pi_mask):
+        axes = res[pi_mask, :3]
```
AlexanderFabisch: The variable name `axes` is very generic. These are not all axes, but the axes for which the rotation is pi. I'd suggest renaming to `axes_with_pi_rotation`.

pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -45,6 +45,19 @@ def norm_axis_angles(a):
...
+        is_zero = np.isclose(axes, 0.0)
```
AlexanderFabisch: I am not entirely sure whether we want to use `np.isclose` here. At least the default threshold might not be appropriate. Which threshold does it use?

pytransform3d/batch_rotations/test/test_batch_rotations.py
```diff
@@ -659,3 +659,18 @@ def test_smooth_quaternion_trajectory_empty():
         ValueError, match=r"At least one quaternion is expected"
     ):
         pbr.smooth_quaternion_trajectory(np.zeros((0, 4)))
+
+
+def test_norm_axis_angle_180_degrees_deterministic_batch():
```
AlexanderFabisch, line 664: Please also add test cases with other array shapes:

- shape (4,)
- shape (N, M, 4)

pytransform3d/rotations/test/test_axis_angle.py
```diff
@@ -284,3 +284,15 @@ def test_mrp_from_axis_angle():
     assert_array_almost_equal(
         [1.0, 0.0, 0.0], pr.mrp_from_axis_angle([1.0, 0.0, 0.0, np.pi])
     )
+
+
+def test_norm_axis_angle_180_degrees_deterministic():
```
AlexanderFabisch, line 289: Same here, please add tests for [0,0,0] and random axes.

pytransform3d/rotations/_axis_angle.py
```diff
@@ -94,6 +94,18 @@ def norm_axis_angle(a):
...
+        is_zero = np.isclose(res[:3], 0.0)
+
```
AlexanderFabisch: I wouldn't put a blank line here. The only purpose of `is_zero` is to compute `non_zero_indices`, so they are conceptually connected and, hence, should be so spatially.

AlexanderFabisch (commented):
Here are my comments.

pytransform3d/rotations/_axis_angle.py
TabishShahMohsin: Have now removed it.

pytransform3d/rotations/test/test_axis_angle.py
TabishShahMohsin: Done! I've added many test cases and tried to make them consistent with other test functions.

pytransform3d/batch_rotations/_axis_angle.py
TabishShahMohsin: I noticed batch_rotations/_matrix.py line 51 (which comes from #43 ):
    angle_close_to_pi = np.abs(angles - np.pi) < 1e-4
Also "tolerance=1e-6" is used at many many places, so I have added the same.

pytransform3d/batch_rotations/_axis_angle.py
TabishShahMohsin: I have corrected the axis=1 bug, and changed the variable names.

pytransform3d/batch_rotations/_axis_angle.py
TabishShahMohsin: Had to use this for (4, )

pytransform3d/batch_rotations/_axis_angle.py
TabishShahMohsin: Removed it.

pytransform3d/batch_rotations/test/test_batch_rotations.py
```diff
@@ -659,3 +659,45 @@ def test_smooth_quaternion_trajectory_empty():
...
+    ):
+        assert_array_almost_equal(r, pr.norm_axis_angle(a))
+
+    # random 
+    rng = np.random.default_rng(39232)
+    axes = pbr.norm_vectors(rng.standard_normal(size=(10, 3)))
+    for axis in axes:
+        a_random = np.hstack((axis, [np.pi]))
+        res_random = pbr.norm_axis_angles(a_random)
+        assert_array_almost_equal(res_random, pr.norm_axis_angle(a_random))
```
AlexanderFabisch, line 702: What if we flip the axis signs? Does it still produce the same axis consistently?

pytransform3d/batch_rotations/test/test_batch_rotations.py
```diff
@@ -659,3 +659,45 @@ def test_smooth_quaternion_trajectory_empty():
...
+
+    # [0, 0, 0] 
+    a_zero_axis = np.array([0.0, 0.0, 0.0, np.pi])
+    res_zero_axis = pbr.norm_axis_angles(a_zero_axis)
+    assert_array_almost_equal(res_zero_axis, pr.norm_axis_angle(a_zero_axis))
```
AlexanderFabisch: The test does not make a lot of sense. Which axis-angle representation do we expect here exactly?

pytransform3d/batch_rotations/test/test_batch_rotations.py
```diff
@@ -659,3 +659,45 @@ def test_smooth_quaternion_trajectory_empty():
         ValueError, match=r"At least one quaternion is expected"
     ):
         pbr.smooth_quaternion_trajectory(np.zeros((0, 4)))
+
+
+def test_norm_axis_angle_180_degrees_deterministic_batch():
+    # (4,)
+    a_single = np.array([-1.0, 0.0, 0.0, np.pi])
+    res_single = pbr.norm_axis_angles(a_single)
+    assert_array_almost_equal(res_single, pr.norm_axis_angle(a_single))
```
AlexanderFabisch: Which result do we expect here?

pytransform3d/batch_rotations/test/test_batch_rotations.py
AlexanderFabisch: The docstring says [1, 0, 0, 0]. Is it the case here?

pytransform3d/rotations/_axis_angle.py
```diff
@@ -64,20 +64,28 @@ def check_compact_axis_angle(a):
...
+    tolerance : float
+        Tolerance for checking if the angle is close to pi and if
+        components of the axis are close to zero, used to make the axis
+        deterministic for 180 degree rotations.
+
     Returns
     -------
     a : array, shape (4,)
         Axis of rotation and rotation angle: (x, y, z, angle). The length
         of the axis vector is 1 and the angle is in [0, pi). No rotation
```
AlexanderFabisch: Please correct the range to [0, pi] as we now return angles pi explicitly.

pytransform3d/batch_rotations/_axis_angle.py
AlexanderFabisch: Thanks. Could you check [0, 0, 0, pi] explicitly? Does it return [1, 0, 0, 0] as suggested in the docstring?

pytransform3d/batch_rotations/_axis_angle.py
AlexanderFabisch: But does it work for, e.g., 3d arrays? Would `first_non_zero_axis_components = axes_with_pi_rotation[..., first_non_zero_idx_per_entry]` work? (Maybe not.) I just feel there should be a simpler solution than `arange`...

pytransform3d/batch_rotations/_axis_angle.py
AlexanderFabisch: I think this matches too broadly. There is still valuable information in the sign of the rotation axis when the angle is within the range of `1e-4` or `1e-6` of pi. This might be critical in applications, in which high numerical precision is required.

`1e-4` was specifically selected to bound the error of that conversion function. We would have to do a similar analysis for this case: measure the minimum of the angle between the normalized axes before and after normalization and abs(this value - pi) (or something similar). However, I think it is perfectly safe to just check angle == pi.

pytransform3d/batch_rotations/test/test_batch_rotations.py
TabishShahMohsin: Yes

pytransform3d/batch_rotations/test/test_batch_rotations.py
TabishShahMohsin: [1.0, 0.0, 0.0, np.pi]

TabishShahMohsin:
Thanks for the thorough review and patience.

pytransform3d/batch_rotations/test/test_batch_rotations.py
AlexanderFabisch: Then please state this explicitly in the test. It is currently a circular test.

pytransform3d/batch_rotations/test/test_batch_rotations.py
AlexanderFabisch: Then please write this explicitly. It is currently a circular test.

pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -45,6 +53,18 @@ def norm_axis_angles(a):
 
     res[rot_mask, 3] = angle_normalized[rot_mask]
 
+    # Issue #366: Make axis deterministic for 180 degree rotations
+    # the first non-zero component of axis should be positive.
+    pi_mask = (np.abs(res[..., 3] - np.pi) < tolerance) & rot_mask
```
AlexanderFabisch, line 58: Checking `res[..., 3] == np.pi` seems to be the most reasonable solution. There seems to be no good argument for a broader range defined by `threshold`.

pytransform3d/rotations/_axis_angle.py
```diff
@@ -94,6 +102,17 @@ def norm_axis_angle(a):
 
     res[3] = angle
 
+    # Issue #366: Make axis deterministic for 180 degree rotations
+    # the first non-zero component of axis should be positive.
+    if np.abs(angle - np.pi) < tolerance:
```
AlexanderFabisch, line 107: Checking `angle == np.pi` seems to be the most reasonable solution. There seems to be no good argument for a broader range defined by `threshold`.

pytransform3d/batch_rotations/test/test_batch_rotations.py
AlexanderFabisch: Please add this check.

pytransform3d/batch_rotations/test/test_batch_rotations.py
```diff
@@ -659,3 +659,53 @@ def test_smooth_quaternion_trajectory_empty():
...
+    # random 
+    rng = np.random.default_rng(39232)
+    axes = pbr.norm_vectors(rng.standard_normal(size=(10, 3)))
+    for axis in axes:
+        a_random = np.hstack((axis, [np.pi]))
+        res_random = pbr.norm_axis_angles(a_random)
+        assert_array_almost_equal(res_random, pr.norm_axis_angle(a_random))
+        a_random = np.hstack((-axis, [np.pi]))
+        res_random = pbr.norm_axis_angles(a_random)
+        assert_array_almost_equal(res_random, pr.norm_axis_angle(a_random))
```
AlexanderFabisch, line 705: This test is still circular. I'd convert both the unnormalized and the normalized axis-angle representation to quaternion or rotation matrix and check whether they are identical.

pytransform3d/rotations/_axis_angle.py
TabishShahMohsin: Exact floating comparison is fragile. It can break even on mathematically equivalent expressions.
```python
pi = np.pi 
for i in range(1, 50):
  a = (pi / i) * i
  if a!=pi:
    print(i)
```
This prints 25 and 49.

pytransform3d/batch_rotations/test/test_batch_rotations.py
TabishShahMohsin: I though that it so to make it consistent with the existing tests, have added the test.

pytransform3d/rotations/_axis_angle.py
AlexanderFabisch: I believe this is more a problem of the previous processing steps, not something that we have to or can fix here.

The problem is: When the computation is exact, accepting a broader range at this check messes up the result. It might result in a slightly wrong axis.

pytransform3d/batch_rotations/test/test_batch_rotations.py
```diff
@@ -659,3 +659,58 @@ def test_smooth_quaternion_trajectory_empty():
...
+    # random 
+    rng = np.random.default_rng(39232)
+    axes = pbr.norm_vectors(rng.standard_normal(size=(10, 3)))
+    for axis in axes:
+        for sign in [1.0, -1.0]:
+            a_random = np.hstack((sign * axis, [np.pi]))
+            res_random = pbr.norm_axis_angles(a_random)
+
+            # scalar vs batch consistency check
+            assert_array_almost_equal(res_random, pr.norm_axis_angle(a_random))
```
AlexanderFabisch: This check still does not make any sense to me. Both the expected and the actual value go through exactly the same function.

pytransform3d/batch_rotations/test/test_batch_rotations.py
TabishShahMohsin: I'm comparing the multi-dim batch function (pbr.norm_axis_angles) to the scalar one (pr.norm_axis_angle), just like the other tests in test_batch_rotations.py.

pytransform3d/rotations/_axis_angle.py
```diff
@@ -94,6 +97,17 @@ def norm_axis_angle(a):
 
     res[3] = angle
 
+    # Issue #366: Make axis deterministic for 180 degree rotations
+    # the first non-zero component of axis should be positive.
+    if angle == np.pi:
+        is_zero = res[:3] == 0.0
+        non_zero_indices = np.where(~is_zero)[0]
+        
+        if len(non_zero_indices) > 0:
```
AlexanderFabisch: This seems to be never False in the unit tests. Could you add a test case for this branch?

pytransform3d/rotations/_axis_angle.py
```diff
@@ -94,6 +97,17 @@ def norm_axis_angle(a):
 
     res[3] = angle
 
+    # Issue #366: Make axis deterministic for 180 degree rotations
+    # the first non-zero component of axis should be positive.
+    if angle == np.pi:
+        is_zero = res[:3] == 0.0
+        non_zero_indices = np.where(~is_zero)[0]
```
AlexanderFabisch:
```suggestion
        nonzero_axis_components = res[:3] != 0.0
        non_zero_indices = np.where(nonzero_axis_components)[0]
```

pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -45,6 +48,18 @@ def norm_axis_angles(a):
 
     res[rot_mask, 3] = angle_normalized[rot_mask]
 
+    # Issue #366: Make axis deterministic for 180 degree rotations
+    # the first non-zero component of axis should be positive.
+    pi_mask = (res[..., 3] == np.pi) & rot_mask
+    if np.any(pi_mask):
+        axes_with_pi_rotation = res[pi_mask, :3]
+        is_zero = axes_with_pi_rotation == 0.0
+        first_non_zero_idx_per_entry = np.argmax(~is_zero, axis=-1)
```
AlexanderFabisch:
```suggestion
        nonzero_axis_components = axes_with_pi_rotation != 0.0
        first_non_zero_idx_per_entry = np.argmax(nonzero_axis_components, axis=-1)
```

pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -45,6 +48,18 @@ def norm_axis_angles(a):
...
+        row_indices = np.arange(len(axes_with_pi_rotation))
+        first_non_zero_axis_components = axes_with_pi_rotation[row_indices, first_non_zero_idx_per_entry]
```
AlexanderFabisch: Lines are too long here. Please apply black and ruff in the root folder of the pytransform3d repo:

```
black . --check
ruff check --output-format=full .
```

AlexanderFabisch:
```suggestion
        first_non_zero_axis_components = np.take_along_axis(
            axes_with_pi_rotation,
            first_non_zero_idx_per_entry[:, np.newaxis],
            axis=-1,
        ).squeeze(axis=-1)
```

AlexanderFabisch:
Thanks for your patience with my nitpicking @TabishShahMohsin . I want to make this code as elegant and correct as possible. I added some additional suggestions.

PR #373, opened 2026-09-10
Fix batch axis-angle conversion without an output buffer

Doribelove:
`matrices_from_compact_axis_angles` raises `AttributeError` when called with precomputed axes and angles without an output buffer, or with compact axis-angle inputs supplied as Python lists. Non-axis-aligned integer compact coordinates also produce incorrect matrices because their normalized axes are stored in an integer array.

For example, this documented parameter combination should return a 90-degree rotation about z, but currently tries to access `None.shape`:

```python
import numpy as np
from pytransform3d.batch_rotations import matrices_from_compact_axis_angles

matrices_from_compact_axis_angles(
    axes=np.array([0.0, 0.0, 1.0]), angles=np.pi / 2
)
```

Allocate the output from the already computed broadcast shape, `uxs.shape`. When both axes and angles are supplied, the shape follows those parameters even if an unused `A` argument is also present. Missing `A` together with incomplete precomputed parameters now raises a descriptive `ValueError`, and the docstring marks axes and angles as optional.

Boolean and integer compact coordinates are converted to floating point before axis normalization. Existing floating-input calculations and supplied output-buffer behavior are preserved. Tests include integer lists `[[0, 0, 1]]`, `[[0, 1, 1]]`, and the zero vector, as well as the missing-parameter cases, batches, empty batches, broadcasting, and explicit output buffers. Matrix values are checked against scalar rotation conversions.

Validation on `develop` base `cf7393f47cbf7111badb1e96c9be14402ffe5787`, using Python 3.10.12 and NumPy 2.2.6:

- Initial regression tests with the original upstream implementation: **8 failed, 53 passed** in the batch-rotation test file.
- Review regressions against the first PR commit: **5 failed, 62 passed**. These expose the incomplete-argument cases, the non-axis-aligned integer input, and the requested output-shape change.
- Full package tests with the final revision: **3,094 passed, 3 skipped**. The three skipped mesh-loader tests require optional Open3D.
- Separate checks confirm the integer non-axis-aligned case now agrees with the scalar reference, and float16/float32/float64/longdouble calculations and strided output buffers retain their prior behavior.
- Ruff, Black, and `git diff --check` passed. Wheel and sdist builds passed for the initial submission; they have not been repeated for this review revision.

AI disclosure: Codex assisted with source investigation, implementation, tests, this description, and the review revisions. Validation was executed locally.

### DIFF
--- a/pytransform3d/batch_rotations/_axis_angle.py
+++ b/pytransform3d/batch_rotations/_axis_angle.py
@@ -63,11 +63,11 @@ def matrices_from_compact_axis_angles(A=None, axes=None, angles=None, out=None):
         Axes of rotation and rotation angles in compact representation:
         angle * (x, y, z). If omitted, both axes and angles must be provided.
 
-    axes : array, shape (..., 3)
+    axes : array, shape (..., 3), optional (default: None)
         If the unit axes of rotation have been precomputed, you can pass them
         here.
 
-    angles : array, shape (...)
+    angles : array, shape (...), optional (default: None)
         If the angles have been precomputed, you can pass them here.
 
     out : array, shape (..., 3, 3), optional (default: new array)
@@ -77,13 +77,24 @@ def matrices_from_compact_axis_angles(A=None, axes=None, angles=None, out=None):
     -------
     Rs : array, shape (..., 3, 3)
         Rotation matrices
+
+    Raises
+    ------
+    ValueError
+        If A is omitted and axes or angles are not provided.
     """
+    if A is None and (axes is None or angles is None):
+        raise ValueError("Either A or both axes and angles must be provided.")
+
     if angles is None:
         thetas = np.linalg.norm(A, axis=-1)
     else:
         thetas = np.asarray(angles)
 
     if axes is None:
+        A = np.asarray(A)
+        if A.dtype.kind in "biu":
+            A = A.astype(float)
         omega_unit = norm_vectors(A)
     else:
         omega_unit = axes
@@ -105,9 +116,7 @@ def matrices_from_compact_axis_angles(A=None, axes=None, angles=None, out=None):
     ciuyuz = ciuy * uz
 
     if out is None:
-        # Without A, use the broadcast shape of the axes and angles.
-        shape = np.shape(A)[:-1] if A is not None else uxs.shape
-        out = np.empty(shape + (3, 3))
+        out = np.empty(uxs.shape + (3, 3))
 
     out[..., 0, 0] = ciux * ux + c
     out[..., 0, 1] = ciuxuy - uzs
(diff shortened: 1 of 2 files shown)


pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -105,7 +105,9 @@ def matrices_from_compact_axis_angles(A=None, axes=None, angles=None, out=None):
     ciuyuz = ciuy * uz
 
     if out is None:
-        out = np.empty(A.shape[:-1] + (3, 3))
+        # Without A, use the broadcast shape of the axes and angles.
+        shape = np.shape(A)[:-1] if A is not None else uxs.shape
```
AlexanderFabisch: Please use `uxs.shape` directly. I don't see a reason why we should use another way to determine the shape when A is given.

pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -59,9 +59,9 @@ def matrices_from_compact_axis_angles(A=None, axes=None, angles=None, out=None):
 
     Parameters
     ----------
-    A : array-like, shape (..., 3)
+    A : array-like, shape (..., 3), optional (default: None)
         Axes of rotation and rotation angles in compact representation:
-        angle * (x, y, z)
+        angle * (x, y, z). If omitted, both axes and angles must be provided.
```
AlexanderFabisch, line 64: > If omitted, both axes and angles must be provided.

Can we check this and raise a ValueError if this is not the case?

pytransform3d/batch_rotations/test/test_batch_rotations.py
```diff
@@ -329,6 +329,67 @@ def test_quaternions_from_matrices_4d():
         pr.assert_quaternion_equal(q, q2[1, 1])
 
 
+@pytest.mark.parametrize("shape", [(), (3,), (2, 3), (0,), (2, 0)])
```
AlexanderFabisch, line 332: Can we also add a test for [[0, 0, 1]]? This is an integer list and will fail.

pytransform3d/batch_rotations/_axis_angle.py
```diff
@@ -59,9 +59,9 @@ def matrices_from_compact_axis_angles(A=None, axes=None, angles=None, out=None):
...
 
     axes : array, shape (..., 3)
```
AlexanderFabisch: axes and angles are also optional, right? Could we emphasize this with `, optional`?

AlexanderFabisch:
Hi @Doribelove , thanks for the contribution! I added a few comments to the PR.

AlexanderFabisch:
Merged. Thanks for the contribution!

